PR #20, opened 2020-10-11
feat: added type package

kyranet:


packages/type/src/index.ts
```diff
@@ -0,0 +1,156 @@
...
+		  process.binding('util');
+
+/**
+ * The class for deep checking Types
+ */
+export class Type {
+	/**
+	 * The value to generate a deep Type of
+	 */
+	public value: unknown;
```
vladfrangu, line 17: Isn't this technically readonly?

vladfrangu (dismissed):

packages/type/src/index.ts
kyranet: Fixed in ce4ddd0.

Favna (approved):

vladfrangu (approved):

cfanoulis (approved):

PR #21, opened 2020-10-11
fix(utilities): fixed up mergeDefault

Favna:


packages/utilities/src/lib/mergeDefault.ts
```diff
@@ -1,27 +1,61 @@
...
+	if (!overwrites) return deepClone(base) as DeepRequired<A & B>;
+
+	for (const [key, value] of Object.entries(base)) {
+		const givenValue = Reflect.get(overwrites, key);
 		if (typeof givenValue === 'undefined') {
-			Reflect.set(given, key, deepClone(value));
+			Reflect.set(overwrites, key, deepClone(value));
 		} else if (isObject(givenValue)) {
-			Reflect.set(given, key, mergeDefault(value as NonNullObject, givenValue));
+			Reflect.set(overwrites, key, mergeDefault((value ?? {}) as NonNullObject, givenValue));
```
Favna, line 56: FYI this is the bugfix part

vladfrangu, line 56: Why are we `?? {}`?

cfanoulis (approved):
um like, ok?

vladfrangu (changes requested):

packages/utilities/src/lib/mergeDefault.ts
Favna: image

Favna: Bug that was carried over from Klasa

vladfrangu: I don't know, it will still break if I do
```ts
mergeDefault({ a: 1 }, { a: { b: true } })
```
won't it?

Favna: it won't

vladfrangu (approved):

kyranet (approved):

PR #28, opened 2020-11-04
Various fixes and chores

Favna:
**rebase this, don't squash it**

- Removed hacktoberfest stuff
- Fixed docgen
- Fixed eslint config to actually be properly compatible with prettier. Previously it was only compatible when also re-extending prettier config in the packages using @sapphire/eslint-config to re-disable some of the rules.


.github/workflows/pr-labeler.yml
```diff
@@ -1,14 +0,0 @@
...
-
-jobs:
-    pr_labeler:
-        name: Automatic Pull Request Labeler
-        runs-on: ubuntu-latest
-        steps:
-            - uses: actions/labeler@main
-              with:
-                  repo-token: '${{ secrets.GITHUB_TOKEN }}'
-                  sync-labels: true
```
vladfrangu, line 14: Don't delete, rename to .disabled or something

vladfrangu (changes requested):

.github/workflows/pr-labeler.yml
Favna: Yeah kinda too late, it's been removed in all the other repos already as well (Framework and pieces were approved by @kyranet)

vladfrangu: ree, fineee

vladfrangu (approved):

kyranet (approved):

PR #35, opened 2020-11-18
feat(time-utilities): add defaults for DurationFormatter

Soumil07:
It's sensible to have DurationFormatter default to en-US, since most users will not require language support.


packages/time-utilities/src/lib/constants.ts
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { DurationFormatAssetsTime, TimeTypes } from '..';
```
Favna, line 2: Can you change this to the file(s) where these imports actually are? I'm not much a fan of importing from `src/index.ts` in other files.

Favna (changes requested):

packages/time-utilities/src/lib/constants.ts
Soumil07: Resolved in 38c91ba

Favna (approved):

PR #50, opened 2020-12-25
fix(time-utilities): make displayUTC's time argument optional

kyranet:


packages/time-utilities/src/lib/Timestamp.ts
```diff
@@ -158,7 +158,7 @@ export class Timestamp {
 	 * @since 1.0.0
 	 * @param time The time to display in utc
 	 */
-	public displayUTC(time: TimeResolvable): string {
+	public displayUTC(time: TimeResolvable = new Date()): string {
```
vladfrangu, line 161: Redundant as `Timestamp.utc` already sets it to new Date
```suggestion
	public displayUTC(time?: TimeResolvable): string {
```

vladfrangu (changes requested):

vladfrangu (approved):

PR #57, opened 2021-01-12
feat(discord.js-utilities): add checking index changes and documentation

RealShadowNova:
This PR implements a feature `disableResponseEdit` to the `IPaginatedMessageAction`. This enables developers to have an action only execute without editing the page to the current index.

This PR also includes documentation for the utility as a whole. Some of which was not written the best, so if you think something should be different go right ahead and suggest it.

### DIFF
--- a/packages/discord.js-utilities/src/lib/PaginatedMessage.ts
+++ b/packages/discord.js-utilities/src/lib/PaginatedMessage.ts
@@ -18,6 +18,7 @@ import type { APIMessage, User, TextChannel, NewsChannel, Message, MessageReacti
  * @example
  * ```typescript
  * // To utilize actions you can use the IPaginatedMessageAction by implementing it into a class.
+ * // [[PaginatedMessage]] requires you to have the class initalized.
  *
  * class ForwardAction implements IPaginatedMessageAction {
  *   public id = '▶️';
--- a/packages/discord.js-utilities/src/lib/PaginatedMessage.ts
+++ b/packages/discord.js-utilities/src/lib/PaginatedMessage.ts
@@ -18,7 +18,7 @@ import type { APIMessage, User, TextChannel, NewsChannel, Message, MessageReacti
  * @example
  * ```typescript
  * // To utilize actions you can use the IPaginatedMessageAction by implementing it into a class.
- * // [[PaginatedMessage]] requires you to have the class initalized.
+ * // [[PaginatedMessage]] requires you to have the class initialized using `new`.
  *
  * class ForwardAction implements IPaginatedMessageAction {
  *   public id = '▶️';
--- a/packages/discord.js-utilities/src/lib/PaginatedMessage.ts
+++ b/packages/discord.js-utilities/src/lib/PaginatedMessage.ts
@@ -1,23 +1,23 @@
 import type { APIMessage, User, TextChannel, NewsChannel, Message, MessageReaction, ReactionCollector } from 'discord.js';
 
 /**
- * This is a PaginatedMessage, a very powerful tool to paginate messages (usually embeds).
- * The developer must either use this class directly or extend it.
+ * This is a [[PaginatedMessage]], a utility to paginate messages (usually embeds).
+ * You must either use this class directly or extend it.
  *
- * PaginatedMessage uses actions, these are essentially reaction emojis, when triggered run the said action.
- * The developer can utilize their own actions or can use the [[PaginatedMessage.defaultActions]].
+ * [[PaginatedMessage]] uses actions, these are essentially reaction emojis, when triggered run the said action.
+ * You can utilize your own actions, or you can use the [[PaginatedMessage.defaultActions]].
  * [[PaginatedMessage.defaultActions]] is also static so you can modify these directly.
  *
- * PaginatedMessage also uses pages, these are simply an {@link https://discord.js.org/#/docs/main/stable/class/APIMessage APIMessage}.
+ * [[PaginatedMessage]] also uses pages, these are simply {@link https://discord.js.org/#/docs/main/stable/class/APIMessage APIMessages}.
  *
  * @example
  * ```typescript
  * const handler = new PaginatedMessage();
- * ````
+ * ```
  *
  * @example
  * ```typescript
- * // To utilize actions you can use the [[IPaginatedMessageAction]] by implementing it into a class.
+ * // To utilize actions you can use the IPaginatedMessageAction by implementing it into a class.
  *
  * class ForwardAction implements IPaginatedMessageAction {
  *   public id = '▶️';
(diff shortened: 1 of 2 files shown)


Favna (dismissed):

packages/discord.js-utilities/src/lib/PaginatedMessage.ts
```diff
@@ -1,65 +1,162 @@
...
 	public index = 0;
 
+	/**
+	 * The amount of time to idle before the paginator is closed. Defaults to `20 * 1000`.
+	 */
 	public idle = 20 * 1000;
 
+	/**
+	 * Constructor for the [[PaginatedMessage]] class
+	 * @param __namedParameters The [[PaginatedMessageOptions]] for this instance of the [[PaginatedMessage]] class
```
kyranet, line 70:
```suggestion
	 * @param options The [[PaginatedMessageOptions]] for this instance of the [[PaginatedMessage]] class
```

packages/discord.js-utilities/src/lib/PaginatedMessage.ts
```diff
@@ -93,30 +190,43 @@ export class PaginatedMessage {
 
 				const page = await this.resolvePage();
 
-				await response.edit(page!);
+				if (!action.disabledResponseEdit) await response.edit(page!);
```
kyranet, line 193: Shouldn't the resolveImage be skipped alongside the response.edit?

packages/discord.js-utilities/src/lib/PaginatedMessage.ts
```diff
@@ -170,11 +281,39 @@ export class PaginatedMessage {
...
+ *   run: ({ response, collector }) => {
+ *     await response!.reactions.removeAll();
+ *     collector!.stop();
+ *   }
+ * }```
+ */
 export interface IPaginatedMessageAction {
 	id: string;
+	disabledResponseEdit?: boolean;
 	run(context: PaginatedMessageActionContext): Awaited<unknown>;
```
kyranet, line 311: To be honest, I'm not a fan of doing this statically, there are two solutions:

- **Semver Minor**: We track the index state, if it changed, after the call, we edit the message. This option is more intuitive and easier to use.
- **Semver Major**: Make the function accept a value that tells whether or not the message should be refreshed.

kyranet (changes requested):

packages/discord.js-utilities/src/lib/PaginatedMessage.ts
Favna: Wrong. TypeDoc named destructured parameters as `__namedParameters`. Using `options` here will cause a mismatch.

### With `options`:
***There is no description for the options***
image


### With `__namedParameters`:
***There is a description for the options***
image

packages/discord.js-utilities/src/lib/PaginatedMessage.ts
RealShadowNova: Ok, I get that I can change that right now.
I could probably implement both tbh.
Would the action's run method return a boolean whether to edit the response or not? If that's the case we would explicitly check the type of the return value to see if it's boolean instead of converting it to a boolean.

kyranet: If you implement both, you'll end up with 2.0.0 instead of 1.1.0 if you implement only once.
Besides, they conflict with each other.

RealShadowNova: Then can't we handle that conflict? A couple conditions on whether the response should be edited:

- Action's boolean is `true` AND index hasn't changed THEN do not edit response.
- Action's boolean is `false` AND index hasn't changed THEN edit response.
- Action's boolean is `true` AND index has changed THEN do not edit response.
- Action's boolean is `false` AND index has changed THEN edit response.
- Action's boolean is invalid AND index hasn't changed THEN do not edit response.
- Action's boolean is invalid AND index has changed THEN edit response.

In other words if the boolean is `true` then do NOT edit the response, otherwise edit.

kyranet: Please don't try to make this API any harder than my Intelligent Behaviour System's practices, keep it simple, just edit when index was changed.

RealShadowNova: Alright, I'll change that.

kyranet (approved):

packages/discord.js-utilities/src/lib/PaginatedMessage.ts
```diff
@@ -1,65 +1,162 @@
...
+ * @example
+ * ```typescript
+ * const handler = new PaginatedMessage();
+ * ```
+ *
+ * @example
+ * ```typescript
+ * // To utilize actions you can use the IPaginatedMessageAction by implementing it into a class.
+ *
+ * class ForwardAction implements IPaginatedMessageAction {
```
vladfrangu, line 22: This presents a class...however, said class would need to be instantiated *somewhere and by someone*

vladfrangu (changes requested):

packages/discord.js-utilities/src/lib/PaginatedMessage.ts
RealShadowNova: Yes, that's correct which is why if you read further in that same code block it states it can also pass an object.

Favna (approved):

packages/discord.js-utilities/src/lib/PaginatedMessage.ts
vladfrangu: Doesn't change the fact you're presenting the possibility of people using raw classes in addAction, which is not handled. Either document that you NEED to pass an instance yourself, or handle it in the code

RealShadowNova: What do you think `IPaginatedMessageAction` is? It's an interface, meaning you can implement it into a class or directly an object. It clearly states this in the documentation. Your reasoning does not make sense imo.

kyranet: `class X implements IPaginatedMessageAction {}` only means that instances of `X` are assignable to the type `IPaginatedMessageAction`, not `X` itself.

packages/discord.js-utilities/src/lib/PaginatedMessage.ts
```diff
@@ -1,65 +1,163 @@
...
+ *
+ * @example
+ * ```typescript
+ * const handler = new PaginatedMessage();
+ * ```
+ *
+ * @example
+ * ```typescript
+ * // To utilize actions you can use the IPaginatedMessageAction by implementing it into a class.
+ * // [[PaginatedMessage]] requires you to have the class initalized.
```
kyranet, line 21:
```suggestion
 * // [[PaginatedMessage]] requires you to have the class initialized using `new`.
```

kyranet (changes requested):

kyranet (approved):

vladfrangu (approved):

Favna (approved):

PR #61, opened 2021-01-18
fix(decorators): remove `client` from `ApplyOptions`'s callback

kyranet:


Favna (changes requested):
Remove callback entirely and reference DI in TSDoc for accessing client

Favna (approved):

PR #62, opened 2021-01-20
feat: added MessageOptions, make PM and LPM work stateless

kyranet:


packages/discord.js-utilities/src/lib/builders/MessageBuilder.ts
```diff
@@ -0,0 +1,168 @@
...
+
+	public constructor(options?: MessageBuilderResolvable) {
+		this.tts = options?.tts;
+		this.nonce = options?.nonce;
+		this.content = options?.content;
+		this.embed = options?.embed;
+		this.allowedMentions = options?.allowedMentions;
+		this.files = options?.files;
+		this.code = options?.code;
+		this.split = options?.split;
```
vladfrangu, line 71: theres no defaults set
also add a default options static object we can use (think always setting allowedmentions to none for example)

vladfrangu (changes requested):

packages/discord.js-utilities/src/lib/builders/MessageBuilder.ts
kyranet: And... what do you propose the default to be? We can't make the default be a MessageBuilder because we need to construct the class itself which would require the static field to be set and constructed...

vladfrangu: Make them messageOptions too! Just set everything to null, or to reasonable defaults (no allowed mentions, no tts, you'll figure it out)

kyranet: How 'bout everything `undefined`, so the defaults are Discord's defaults?

vladfrangu: Deal

kyranet: Fixed in tts
nonce
content
embed
allowedMentions
files
code
split

vladfrangu (approved):

RealShadowNova (approved):

Favna (approved):

PR #75, opened 2021-02-16
fix(discord-utilities): ensure `FormattedCustomEmojiWithGroups` and `ParsedCustomEmojiWithGroups` can be used as substring extraction

Favna:
TSDoc claimed it could be a substring, implementation didn't reflect that.

### DIFF
--- a/packages/discord-utilities/src/lib/regexes.ts
+++ b/packages/discord-utilities/src/lib/regexes.ts
@@ -46,7 +46,7 @@ export const FormattedCustomEmoji = /<a?:\w{2,32}:\d{17,18}>/;
  * @raw `/^(?<animated>a?):(?<name>[^:]+):(?<id>\d{17,19})$/`
  * @remark Capture group 1 can be used to determine whether the emoji is animated or not. It is named `animated`.
  * @remark Capture group 2 is the name of the emoji as it is typed in a message. It is named `name`.
- * @remark Capture group 2 is the ID of the emoji. It is named `id`.
+ * @remark Capture group 3 is the ID of the emoji. It is named `id`.
  */
 export const FormattedCustomEmojiWithGroups = /(?<animated>a?):(?<name>[^:]+):(?<id>\d{17,19})/;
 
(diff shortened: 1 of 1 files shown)


packages/discord-utilities/src/lib/regexes.ts
```diff
@@ -83,7 +83,7 @@ export const ParsedCustomEmoji = /a?:\w{2,32}:\d{17,18}/;
  * @remark Capture group 2 is the name of the emoji as it is typed in a message. It is named `name`.
  * @remark Capture group 2 is the ID of the emoji. It is named `id`.
```
vladfrangu, line 84: 1, 2, 2?

packages/discord-utilities/src/lib/regexes.ts
```diff
@@ -48,7 +48,7 @@ export const FormattedCustomEmoji = /<a?:\w{2,32}:\d{17,18}>/;
...
```
vladfrangu, line 49: 1, 2, 2?

vladfrangu (changes requested):

packages/discord-utilities/src/lib/regexes.ts
Favna: Fixed and checked for other as well

packages/discord-utilities/src/lib/regexes.ts
Favna: Fixed and checked for other as well

kyranet (approved):

packages/discord-utilities/src/lib/regexes.ts
```diff
@@ -81,9 +81,9 @@ export const ParsedCustomEmoji = /a?:\w{2,32}:\d{17,18}/;
  * @raw `/^(?<animated>a?):(?<name>[^:]+):(?<id>\d{17,19})$/`
```
vladfrangu, line 81: raw is wrong :)

vladfrangu (changes requested):

packages/discord-utilities/src/lib/regexes.ts
Favna: Resolved. Also found 2 other raw's that were wrong

vladfrangu (approved):

PR #77, opened 2021-02-23
fix(duration): make invalid input return NaN

kyranet:


vladfrangu (approved):

packages/time-utilities/src/lib/Duration.ts
```diff
@@ -109,11 +110,14 @@ export class Duration {
 			.replace(Duration.kAanRegex, '1')
 			// do math
```
Favna, line 111: who the crap wrote this code

Favna (approved):

PR #79, opened 2021-03-02
fix: remove terser from all packages

Favna:
- fix:  remove terser to allow better debugging & source maps
- chore: refactor all rollup configs to use a base config


packages/async-queue/rollup.config.js
```diff
@@ -0,0 +1,3 @@
+import baseConfig from '../../scripts/rollup.config';
```
kyranet, line 1: Rollup runs ESM?

Favna: Honestly I have no clue what they do. They also support .mjs but when I did that it starting nagging about not being able to find the basic config unless I specifically added .mjs to it as well. Yet as .js it still supports ESM imports... So yeah no idea what they do.

vladfrangu (approved):
LGTM!

kyranet (approved):

PR #94, opened 2021-04-03
fix(discord.js-utilities): improve MessagePrompter typings

noftaly:
`MessagePrompter.run` used to be poorly typed, with `unknown`, which prevented users to use `.catch()` or `.then()`. With a simple generic, it was quite easy to deduce the appropriate return type.
The idea of the interfaces to simplify the return types (rather than having a big nested ternary expression) comes from @kyranet

Because the default strategy of the MessagePrompter is the base strategy, I had to cast the return type. Maybe this can be improved


Nytelife26 (approved):
I didn't even know you could use ternary operators with types until this PR, so actually, thanks for teaching me something along the way. Aside from code noise, looks good, and I'm more about functionality so I'll leave the stylistic decisions to the team.

Favna (approved):
code LGTM. You tested it with a running bot, right?

Favna:
@all-contributors please add @noftaly for code

noftaly:
> code LGTM. You tested it with a running bot, right?

Yes I have:
image

PR #118, opened 2021-05-24
feat: added string-store

kyranet:
### TODO

- [x] Write remaining tests
- [x] Write examples in README
- [x] Write documentation
- [x] Handle edge cases
  - [x] Invalid offset (writing outside buffer length)
  - [x] Invalid offset (reading outside buffer length)
  - [x] Invalid fixed-length array input (not an array, mismatching length)


.github/CODEOWNERS
```diff
@@ -7,6 +7,7 @@ packages/discord.js-utilities/src/lib/*PaginatedMessage* @kyranet @realshadownov
 /packages/ratelimits @kyranet @vladfrangu
 /packages/snowflake @favna @vladfrangu
 /packages/stopwatch @Soumil07 @kyranet @vladfrangu
+/packages/stopwatch @kyranet
```
vladfrangu, line 10: wrong

vladfrangu (changes requested):

packages/string-store/package.json
```diff
@@ -0,0 +1,56 @@
+{
+	"name": "@sapphire/string-store",
+	"version": "0.0.1",
+	"description": "Conveniently store a lot of data into a very small payload.",
```
Nytelife26, line 4:
```suggestion
	"description": "Conveniently store a lot of data in a tiny payload.",
```

packages/string-store/src/lib/data/UnalignedBufferReader.ts
```diff
@@ -0,0 +1,75 @@
+import { Bit, Bit2, Bit4, IUnalignedBuffer } from './IUnalignedBuffer';
+
+export class UnalignedBufferReader extends IUnalignedBuffer {
+	public readBit(): Bit {
+		return this.read(1n) as Bit;
+	}
```
Nytelife26, line 6: Consider using a helper generator instead of re-implementing the same function innumerable times.
This could be done as follows, for instance:

```ts
function readBitX<T>(this: UnalignedBuffer, x: Size) {
    return () => {
        return this.read(x) as T;
    }
}
```
And then applying the generated functions to the class. This is as close to a macro as TypeScript will allow, but more readable and convenient.

packages/string-store/src/lib/data/UnalignedBufferWriter.ts
```diff
@@ -0,0 +1,80 @@
...
+		this.memoryBits += bits;
+		this.ensureCurrentByte();
+	}
+
+	public writeBit(bit: Bit): void {
+		this.memoryData <<= 1n;
+		this.memoryData |= bit & 0b0001n;
+		this.memoryBits += 1n;
+		this.ensureCurrentByte();
+	}
```
Nytelife26, line 19: Same as with `readBit`.

packages/string-store/src/lib/types/Array.ts
```diff
@@ -0,0 +1,40 @@
+import type { IType } from './IType';
+
+export function Array<SType, DType>(type: IType<SType, DType>, length: number | null = null) {
```
Nytelife26, line 3:
```suggestion
export function Array<SType, DType>(type: IType<SType, DType>, length: number | null = null) {
```
It is largely unclear what SType and DType relate to here. Given that `Type` is implied by their usage, ignoring them, it's roughly as vague as `S` and `D`, which appear to have no real standardized use. Consider swapping those out.

.github/workflows/cd-string-store.yml
```diff
@@ -0,0 +1,92 @@
+name: Continuous Delivery - string-store
```
Favna, line 1: This file needs to be removed

packages/string-store/CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+# Change Log
```
Favna, line 1: This file needs to be removed, Lerna auto generates it.

packages/string-store/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+-   Written in TypeScript
+-   Fully tested
+
+## Installation
+
+```sh
+yarn add @sapphire/string-store
+# npm install @sapphire/string-store
+```
```
Favna, line 42: Update to match new template of other packages

packages/string-store/package.json
```diff
@@ -0,0 +1,56 @@
...
+		"type": "git",
+		"url": "git+https://github.com/sapphiredev/utilities.git",
+		"directory": "packages/string-store"
+	},
+	"files": [
+		"dist",
+		"!dist/*.tsbuildinfo"
+	],
+	"engines": {
+		"node": ">=12",
```
Favna, line 37:
```suggestion
		"node": ">=14",
```

Favna (changes requested):

packages/string-store/README.md
```diff
@@ -0,0 +1,124 @@
...
+-   [Installation](#installation)
+-   [Usage](#usage)
+    -   [Basic Usage](#basic-usage)
+-   [API Documentation](#api-documentation)
+-   [Buy us some doughnuts](#buy-us-some-doughnuts)
+-   [Contributors ✨](#contributors-%E2%9C%A8)
+
+## Description
+
+Provides a set of methods and properties that you can use to accurately measure elapsed time.
```
Fyko, line 30: needs to be updated

kyranet: Many things to do, one iOS bug to fix...

Once the iOS bug is fixed, I'll update this PR and proceed to finish StringStore.

packages/string-store/src/lib/StringStoreEntry.ts
```diff
@@ -0,0 +1,46 @@
...
+
+		this.entries.set(name, type);
+		return this;
+	}
+
+	public serialize(writer: UnalignedBufferWriter, ...parameters: readonly unknown[]): void {
+		const i = 0;
+		for (const type of this.entries.values()) {
+			if (i >= parameters.length) writer.writeEmpty(BigInt(type.size));
+			else type.serialize(writer, parameters[i]);
```
favna, line 34:
```suggestion
			else type.serialize(writer, parameters[i]);
			i++;
```

as shown on discord I found a 🐛

packages/string-store/src/lib/StringStoreEntry.ts
```diff
@@ -0,0 +1,46 @@
...
+		if (this.entries.has(name)) {
+			throw new Error(`The entry ${name} already exists.`);
+		}
+
+		this.entries.set(name, type);
+		return this;
+	}
+
+	public serialize(writer: UnalignedBufferWriter, ...parameters: readonly unknown[]): void {
+		const i = 0;
```
favna, line 31:
```suggestion
		let i = 0;
```

packages/string-store/src/lib/data/writer/AlignedWriter.ts
```diff
@@ -0,0 +1,58 @@
...
+	public writeF64(value: number): void {
+		this.offset = this.buffer.writeDoubleLE(value, this.offset);
+	}
+
+	public writeEmpty(bits: number): void {
+		this.offset += Math.ceil(bits / 8);
+	}
+
+	public finish(): void {
+		// NOP: Aligned writer does not use a double buffer.
```
favna, line 52:
```suggestion
		// NO-OP: Aligned writer does not use a double buffer.
```


Not I know but it's "no operation" so it needs 2 O's. No one uses "NOP".

packages/string-store/src/lib/types/Array.ts
favna: Bump, kyra

favna (changes requested):

kyranet:
This PR has been rewritten from the ground up, all reviews from the previous implementation are stale.

Lioness100:
No README examples D:

packages/string-store/package.json
```diff
@@ -0,0 +1,74 @@
+{
+	"name": "@sapphire/string-store",
+	"version": "1.0.0",
+	"description": "High-capacity raw data storage in UTF-16 strings",
+	"author": "@sapphire",
+	"license": "MIT",
+	"main": "dist/cjs/index.cjs",
+	"module": "dist/esm/index.mjs",
+	"browser": "dist/iife/index.global.js",
+	"unpkg": "dist/iife/index.global.js",
```
favna, line 10:
```suggestion
```

Possible removal based on other comments

packages/string-store/package.json
```diff
@@ -0,0 +1,74 @@
...
+	"exports": {
+		"import": {
+			"types": "./dist/esm/index.d.mts",
+			"default": "./dist/esm/index.mjs"
+		},
+		"require": {
+			"types": "./dist/cjs/index.d.cts",
+			"default": "./dist/cjs/index.cjs"
+		},
+		"browser": "./dist/iife/index.global.js"
```
favna, line 21:
```suggestion
		}
```

packages/string-store/package.json
```diff
@@ -0,0 +1,74 @@
...
+		"type": "git",
+		"url": "git+https://github.com/sapphiredev/utilities.git",
+		"directory": "packages/string-store"
+	},
+	"files": [
+		"dist/"
+	],
+	"engines": {
+		"node": ">=v14.0.0",
+		"npm": ">=7.0.0"
```
favna, line 46:
```suggestion
		"node": ">=v20",
		"npm": ">=10"
```

favna (changes requested):
Also needs updating:

- `.github/labels-yml`

1. Is there a way to add a whole object so `new Schema(Id)` is valid somehow? It's a bit weird to be adding `Obj.key` for every single key IMHO

2. Few people will know what all of `int2/4/8/16/32` etc mean, especially those who don't know C-based languages. I count myself among them in fact. Please describe them better.

kyranet:
`.github/labels-yml` already contained that label, and did so since https://github.com/sapphiredev/utilities/commit/db64a22885a13a3e2a6fe2241d816c718707a532

> Is there a way to add a whole object so `new Schema(Id)` is valid somehow? It's a bit weird to be adding `Obj.key` for every single key IMHO

I chose the chainable method since it's a builder in builders, but Zod/Shapeshift-like objects can be done in a follow-up PR, this version is good enough as-is.

> Few people will know what all of `int2/4/8/16/32` etc mean, especially those who don't know C-based languages. I count myself among them in fact. Please describe them better.

`int2` and `int4` don't exist in any language (to my knowledge), they make sense in `string-store` because the unaligned buffer storage makes it take less space than an aligned one (which would store both as an `int8` at minimum).

That aside, I wrote some more documentation, I hope that explains it.

---

By the way, after landing this PR, I'd like to hold off the v1.0.0 release, I want a first version in so I can work on QoL and details without having them around a massive PR, that way they're easier to review and to iterate.

favna (approved):

packages/string-store/README.md
```diff
@@ -0,0 +1,284 @@
...
+npm install @sapphire/string-store
+```
+
+## Usage
+
+**Note**: While this section uses `require`, the imports match 1:1 with ESM imports. For example, `const { SchemaStore } = require('@sapphire/string-store')` is equivalent to `import { SchemaStore } from '@sapphire/string-store'`.
+
+```ts
+// Require the store classes
+const { SchemaStore, Store } = require('@sapphire/string-store');
```
Lioness100, line 42:
```suggestion
const { SchemaStore, Schema } = require('@sapphire/string-store');
```

favna (approved):

vladfrangu (approved):

PR #122, opened 2021-05-31
feat(utilities): update utilities with more stuff

Favna:
- ~~This Awaited is taken from the TS docs: https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#recursive-conditional-types~~
- `NonNullableClass` is backported from [a type from Skyra](https://github.com/skyra-project/skyra/blob/20ea43ab9321576591bef94324b92ef97476fcb2/typescript/src/commands/Tools/topinvites.ts#L79-L81), but made generic

Also [backported from Skyra](https://github.com/skyra-project/skyra/blob/20ea43ab9321576591bef94324b92ef97476fcb2/typescript/src/lib/types/Utils.ts#L29-L46)
- NonNullObject
- AnyObject
- KeyOfType
- Mutable
- StrictRequired
- ArrayElementType


vladfrangu (dismissed):

packages/utilities/src/lib/utilityTypes.ts
```diff
@@ -72,3 +72,91 @@ export type Awaited<T> = PromiseLike<T> | T;
...
+
+/**
+ * An object that is non nullable, to bypass TypeScript not easily working with {@link Record}<{@link PropertyKey}, unknown> in various instances.
+ */
+export interface NonNullObject {}
+
+/**
+ * Any object at all. This can be used if you need to assign the most generic type of object to a type.
+ */
+export type AnyObject = Record<PropertyKey, unknown> | NonNullObject;
```
kyranet, line 92: This type is useless, NonNullObject already includes the record as it's the base.

packages/utilities/src/lib/utilityTypes.ts
```diff
@@ -72,3 +72,91 @@ export type Awaited<T> = PromiseLike<T> | T;
  * Type union for the full 2 billion dollar mistake in the JavaScript ecosystem
  */
 export type Nullish = null | undefined;
+
+/**
+ * Similar to the built in {@link NonNullable}, but properly removes `null` from all keys in the class or interface
+ * This does not recurse deeply, for that use {@link DeepRequired}
+ */
+export type NonNullableClass<T = unknown> = {
```
kyranet, line 80: Maybe we should use NonNullableProperties?

packages/utilities/src/lib/utilityTypes.ts
```diff
@@ -72,3 +72,91 @@ export type Awaited<T> = PromiseLike<T> | T;
...
+ * 	middleName?: string;
+ * 	lastName: string;
+ * 	hobbies: readonly string[];
+ * }
+ *
+ * type BB = KeyOfType<Sample, string>;
+ * // Expected:
+ * // "id" | "lastName"
+ */
+export type KeyOfType<T, V> = {
```
kyranet, line 109: Rename to PropertiesOfType

packages/utilities/src/lib/utilityTypes.ts
```diff
@@ -72,3 +72,91 @@ export type Awaited<T> = PromiseLike<T> | T;
...
+ * type BB = Mutable<Sample>;
+ * // Expected:
+ * // {
+ * //    id: string;
+ * //    hobbies: string[];
+ * // }
+ */
+export type Mutable<T> = {
+	-readonly [P in keyof T]: T[P] extends Array<unknown> | AnyObject | NonNullObject ? Mutable<T[P]> : T[P];
+};
```
kyranet, line 131: This should overload T to make `ReadonlyMap` become `Map`, `ReadonlySet` become `Set`, `readonly T[]` become `T[]`, primitives stay as-is, and objects stripped of readonly.

kyranet (changes requested):
Examples must be inside codeblocks for them to work.

packages/utilities/src/lib/utilityTypes.ts
Favna: what now?

packages/utilities/src/lib/utilityTypes.ts
Favna: Then why is it in Skyra

packages/utilities/src/lib/utilityTypes.ts
Favna: Then why did you name it like this in Skyra. Also in https://github.com/piotrwitek/utility-types it is named [`PickByValue`[(https://github.com/piotrwitek/utility-types#pickbyvaluet-valuetype) which is leagues better than `PropertiesOfType` because this doesn't get the properties of the type, this gets the properties of a type as based on another type.

packages/utilities/src/lib/utilityTypes.ts
Favna: what now?

PR #126, opened 2021-06-10
fix(docs): update-tsdoc-for-vscode-may-2021

Favna:


packages/decorators/src/core-decorators.ts
```diff
@@ -8,7 +8,7 @@ type Ctor<A extends Arr = readonly any[], R = any> = new (...args: A) => R;
  * Decorator function that applies given options to any Sapphire piece
  * @example
  * ```ts
- *	ApplyOptions<CommandOptions>({
+ *	(at)ApplyOptions<CommandOptions>({
```
kyranet, line 11: Is this still broken?

kyranet (changes requested):

packages/decorators/src/core-decorators.ts
Favna: Yeah only decorators are, sadly. :\

kyranet (approved):

PR #128, opened 2021-06-23
feat(discord.js-utilities): verify channel type before removing reaction & allow page prompt customization

kaname-png:
### Changes made
1. permission to manage messages is now optional for the pager. This has no negative effect on the `ReactionCollector` of `discord.js`.
2. It is now possible to customize the `jump prompt message` to a certain page number so that the user can choose a certain message. This is because not all bots are in English language and this allows more customization.

### DIFF
--- a/packages/discord.js-utilities/src/lib/PaginatedMessage.ts
+++ b/packages/discord.js-utilities/src/lib/PaginatedMessage.ts
@@ -77,6 +77,12 @@ export class PaginatedMessage {
 	public idle = 20 * 1000;
 
 	/**
+	 * Custom prompt message when a user wants to jump to a certain page number.
+	 * @default What page would you like to jump to
+	 */
+	public static promptMessage = 'What page would you like to jump to?';
+
+	/**
 	 * Constructor for the {@link PaginatedMessage} class
 	 * @param __namedParameters The {@link PaginatedMessageOptions} for this instance of the {@link PaginatedMessage} class
 	 */
@@ -85,6 +91,11 @@ export class PaginatedMessage {
 
 		for (const page of this.pages) this.messages.push(page instanceof APIMessage ? page : null);
 		for (const action of actions ?? this.constructor.defaultActions) this.actions.set(action.id, action);
+	}
+
+	public setPromptMessage(message: string) {
+		PaginatedMessage.promptMessage = message;
+		return this;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


packages/discord.js-utilities/src/lib/PaginatedMessage.ts
```diff
@@ -272,7 +283,9 @@ export class PaginatedMessage {
 	 * @param user The user that reacted to the message.
 	 */
 	protected async handleCollect(author: User, channel: TextChannel | NewsChannel, reaction: MessageReaction, user: User): Promise<void> {
-		await reaction.users.remove(user);
+		if ((channel as TextChannel).permissionsFor(channel.client.user?.id || '')?.has('MANAGE_MESSAGES')) {
+			await reaction.users.remove(user);
+		}
```
Favna, line 288:
```suggestion
		if (isGuildBasedChannel(channel) && channel.client.user && channel.permissionsFor(channel.client.user.id).has('MANAGE_MESSAGES')) {
			await reaction.users.remove(user);
		}
```

Don't typecast, that's extremely unsafe and this will crash in DMs because `permissionsFor` is not a method on `DMChannel`

packages/discord.js-utilities/src/lib/PaginatedMessage.ts
```diff
@@ -300,7 +313,9 @@ export class PaginatedMessage {
 
 		// Do not remove reactions if the message, channel, or guild, was deleted:
 		if (this.response && !PaginatedMessage.deletionStopReasons.includes(reason)) {
-			await this.response.reactions.removeAll();
+			if ((this.response.channel as TextChannel).permissionsFor(this.response.client.user?.id || '')?.has('MANAGE_MESSAGES')) {
+				await this.response.reactions.removeAll();
+			}
```
Favna, line 318: Similar change as above. Don't typecast, that's extremely unsafe and this will crash in DMs because `permissionsFor` is not a method on `DMChannel`

packages/discord.js-utilities/src/lib/PaginatedMessage.ts
```diff
@@ -355,7 +370,10 @@ export class PaginatedMessage {
 		{
 			id: '⏹️',
 			run: async ({ response, collector }) => {
-				await response.reactions.removeAll();
+				if ((response.channel as TextChannel).permissionsFor(response.client.user?.id || '')?.has('MANAGE_MESSAGES')) {
+					await response.reactions.removeAll();
+				}
+
```
Favna, line 376: Similar change as above. Don't typecast, that's extremely unsafe and this will crash in DMs because `permissionsFor` is not a method on `DMChannel`

Favna (changes requested):

packages/discord.js-utilities/src/lib/PaginatedMessage.ts
kaname-png: done

PR #134, opened 2021-07-17
feat(discord.js-utilities): improve PaginatedMessage

Favna:
### Features

* **discord.js-utilities:** Added TSDoc for many methods
* **discord.js-utilities:** Added caching of messages and handlers to improve performance
* **discord.js-utilities:** Added `addPageBuilder` helper method to make adding pages easier
* **discord.js-utilities:** Added `addPageContent` helper method to make adding pages easier
* **discord.js-utilities:** Added `addPageEmbed` helper method to make adding pages easier
* **discord.js-utilities:** Added `addAsyncPageEmbed` helper method to make adding pages easier. This method is unique from `addPageEmbed` in that the callback can return a `Promise`.
* **discord.js-utilities:** Added `addAsyncPageBuilder` helper method to make adding pages easier. This method is unique from `addPageBuilder` in that the callback can return a `Promise`.
* **discord.js-utilities:** Added `PaginatedMessageOptions.template` to create template `MessageOptions` that apply to all pages
* **discord.js-utilities:** Refactored `PaginatedMessage.setUpReactions` to only add reactions if more than 1 page was added
* **discord.js-utilities:** Refactored to use `@sapphire/time-utilities` and `@sapphire/utilities` instead of duplicating code

### ⚠ BREAKING CHANGES

* **discord.js-utilities:** `PaginatedMessageOptions.run` now takes a single parameter of `Message` instead of 2 parameters (`User` and `TextChannel | NewsChannel`)
* **discord.js-utilities:** Reactions will no longer be added if your `PaginatedMessage` only has 1 page
* **discord.js-utilities:** TypeScript types for various methods that previously had `TextChannel | NewsChannel` have been changed to `Message['channel']`
* **discord.js-utilities:** 1 user can no longer have more than 1 `PaginatedMessage` running. The older one will automatically be cancelled. You can override this by overriding the `run` method.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
 		"@typescript-eslint/eslint-plugin": "^4.28.2",
 		"@typescript-eslint/parser": "^4.28.0",
 		"cz-conventional-changelog": "^3.3.0",
+		"discord-api-types": "^0.18.1",
 		"discord.js": "^12.5.3",
 		"eslint": "^7.30.0",
 		"eslint-config-prettier": "^8.3.0",
--- a/packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
+++ b/packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
@@ -1,16 +1,20 @@
+import { Time } from '@sapphire/time-utilities';
+import { Awaited, isFunction } from '@sapphire/utilities';
+import type { RESTPostAPIChannelMessageJSONBody } from 'discord-api-types/v8';
 import {
 	APIMessage,
 	Collection,
 	Message,
+	MessageEmbed,
+	MessageEmbedOptions,
 	MessageOptions,
 	MessageReaction,
-	NewsChannel,
 	ReactionCollector,
 	Snowflake,
-	TextChannel,
 	User
 } from 'discord.js';
-import { isGuildBasedChannel } from './type-guards';
+import { MessageBuilder } from '../builders/MessageBuilder';
+import { isGuildBasedChannel } from '../type-guards';
 
 /**
  * This is a {@link PaginatedMessage}, a utility to paginate messages (usually embeds).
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
 		"@typescript-eslint/eslint-plugin": "^4.28.2",
 		"@typescript-eslint/parser": "^4.28.0",
 		"cz-conventional-changelog": "^3.3.0",
+		"discord-api-types": "^0.18.1",
```
vladfrangu, line 39: Why is this at the global level?

Favna:
@sapphiredev pack

kyranet (dismissed):

packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
```diff
@@ -178,6 +196,119 @@ export class PaginatedMessage {
...
+	 *
+	 * 		embed
+	 * 			.setColor('#FF0000')
+	 * 			.setDescription(someRemoteData.data);
+	 *
+	 * 		return embed;
+	 * });
+	 * ```
+	 */
+	public addAsyncPageEmbed(cb: MessageEmbed | ((builder: MessageEmbed) => Promise<MessageEmbed>)): this {
```
vladfrangu, line 308: Why just embed? Why not content too, or hell, anything else

packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
```diff
@@ -404,17 +603,40 @@ export class PaginatedMessage {
 	 * @default "What page would you like to jump to?"
 	 */
 	public static promptMessage = 'What page would you like to jump to?';
+
+	/**
+	 * The messages that are currently being handled by a {@link PaginatedMessage}
+	 * The key is the ID of the message that triggered this {@link PaginatedMessage}
+	 *
+	 * This is to ensure that only 1 {@link PaginatedMessage} can run on a specified message at once.
+	 * This is import when having an editable commands solution.
```
vladfrangu, line 612: This. is. IMPORT.

vladfrangu (changes requested):

packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
kyranet: image

package.json
Favna: Because all dev deps are. It's a sneaky way to present a cleaner bundle on npm.

packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
Favna: That's what `addPageBuilder` and `addPageContent` are for. The former doing everything from `MessageOptions`.

But for the actual reason, I only needed this in Skyra and this is a backport after all. I'll add `addAsyncPageBuilder`. There is no point in `addAsyncPageContent` because `addPageContent` doesn't take a callback.

Favna:
@sapphiredev pack

Favna:
@sapphiredev pack

vladfrangu (approved):

packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
```diff
@@ -178,6 +196,142 @@ export class PaginatedMessage {
...
+	 * 			.setDescription(someRemoteData.data);
+	 *
+	 * 		return builder
+	 * 			.setContent('example content')
+	 * 			.setEmbed(embed);
+	 * });
+	 * ```
+	 */
+	public addAsyncPageBuilder(builder: MessageBuilder | ((builder: MessageBuilder) => Promise<MessageBuilder>)): this {
+		return this.addPage(async () => (isFunction(builder) ? builder(new MessageBuilder()) : builder));
```
kyranet, line 262: This isn't using await at all.

kyranet: And why should it be separated? All pages are awaited, regardless of what they accept. In fact, this is redundant.

kyranet (changes requested):

PR #149, opened 2021-08-06
feat(fetch): migrate to using cross-fetch

Favna:
All in all pretty poggers that it can be used in browsers after this

**_Once this is merged then I'll release all utilities packages which includes those for DJS v13._**

BREAKING CHANGE: `fetch` is no longer on the default export. Use a named import instead.
BREAKING CHANGE: This package no longer depends on `node-fetch`
BREAKING CHANGE: This package no longer depends on `@types/node-fetch` for TypeScript users

### DIFF
--- a/packages/fetch/package.json
+++ b/packages/fetch/package.json
@@ -6,6 +6,8 @@
 	"license": "MIT",
 	"main": "dist/index.js",
 	"module": "dist/index.mjs",
+	"browser": "dist/index.umd.js",
+	"unpkg": "dist/index.umd.js",
 	"types": "dist/index.d.ts",
 	"exports": {
 		"import": "./dist/index.mjs",
@@ -16,14 +18,12 @@
 	"scripts": {
 		"test": "jest",
 		"lint": "eslint src tests --ext ts --fix -c ../../.eslintrc",
-		"build": "tsc -b src",
-		"postbuild": "gen-esm-wrapper dist/index.js dist/index.mjs",
+		"build": "rollup -c",
 		"start": "yarn build -w",
 		"prepublishOnly": "yarn build"
 	},
 	"dependencies": {
-		"node-fetch": "^2.6.1",
-		"tslib": "2.x"
+		"cross-fetch": "^3.1.4"
 	},
 	"repository": {
 		"type": "git",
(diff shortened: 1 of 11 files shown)


packages/fetch/package.json
```diff
@@ -16,14 +18,12 @@
...
 		"lint": "eslint src tests --ext ts --fix -c ../../.eslintrc",
-		"build": "tsc -b src",
-		"postbuild": "gen-esm-wrapper dist/index.js dist/index.mjs",
+		"build": "rollup -c",
 		"start": "yarn build -w",
 		"prepublishOnly": "yarn build"
 	},
 	"dependencies": {
-		"node-fetch": "^2.6.1",
-		"tslib": "2.x"
```
vladfrangu, line 26: You sure this isn't needed anymore?

vladfrangu (changes requested):

RealShadowNova (approved):

packages/fetch/package.json
Favna: Yeah, node-fetch is a dep of cross-fetch (package.json: https://cdn.jsdelivr.net/npm/cross-fetch/package.json). And tslib is not needed anymore because the package is now built with Rollup so there is an UMD browser bundle.

That said, it's a good remark you made because I started testing with the UMD bundle and I had to make some more changes to the bundling. I now tested it with this HTML
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Document</title>
    <script type="text/javascript" src="./index.umd.js"></script>
    <script>
        async function sampleData() {
            const data = await SapphireFetch.fetch('http://127.0.0.1:8081', SapphireFetch.FetchResultTypes.Text);
            console.log(data);
        }
    </script>
</head>
<body>
    Hello World!

    <script defer>
        sampleData();
    </script>
</body>
</html>
```

And that works. If I open this HTML on `http://127.0.0.1:8081` then I see a successful fetch in the console returning this same HTML.

vladfrangu (approved):

kyranet (approved):

PR #159, opened 2021-08-23
fix(discord.js-utilities): remove `MessagePrompterStrategies` in favour of `keyof StrategyReturns`

Favna:
fix(discord.js-utilities): remove `MessagePrompterStrategies` in favour of `keyof StrategyReturns`

BREAKING CHANGE: `MessagePrompterStrategies` does no exist as it was not mutable with type augmentation
BREAKING CHANGE: If you had custom strategies then be sure to instead module augment `StrategyReturns` with your added keys.


Favna:
@sapphiredev pack

kyranet (approved):

PR #165, opened 2021-09-11
fix(discord.js-utilities): account thread channels in permission checks

kyranet:
Although I'm not 100% sure if this logic is correct, cc: @vladfrangu

### DIFF
--- a/packages/discord.js-utilities/src/lib/type-guards.ts
+++ b/packages/discord.js-utilities/src/lib/type-guards.ts
@@ -26,7 +26,7 @@ export function isCategoryChannel(channel: ChannelTypes | Nullish): channel is C
  * Checks whether a given channel is a {@link DMChannel}
  * @param channel The channel to check
  */
-export function isDMChannel(channel: ChannelTypes | Nullish): channel is DMChannel {
+export function isDMChannel(channel: ChannelTypes | Nullish): channel is DMChannel | PartialDMChannel {
 	return channel?.type === 'DM';
 }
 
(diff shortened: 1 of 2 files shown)


packages/discord.js-utilities/src/lib/utilities.ts
```diff
@@ -1,5 +1,6 @@
 import { isNullish, Nullish } from '@sapphire/utilities';
 import { Permissions } from 'discord.js';
+import { isThreadChannel } from '..';
```
vladfrangu, line 3: Eek. This import makes me die inside (and I've been avoiding them because I'd rather they be explicit.. idk what you want tho)

vladfrangu (approved):
Overall this looks fine, I don't exactly remember the conditions for archived and reacts (if the client unarchives the thread then reacts or not)

packages/discord.js-utilities/src/lib/type-guards.ts
```diff
@@ -26,7 +26,7 @@ export function isCategoryChannel(channel: ChannelTypes | Nullish): channel is C
  * Checks whether a given channel is a {@link DMChannel}
  * @param channel The channel to check
  */
-export function isDMChannel(channel: ChannelTypes | Nullish): channel is DMChannel {
+export function isDMChannel(channel: ChannelTypes | Nullish): channel is DMChannel | PartialDMChannel {
```
Favna, line 29: In what instance does a bot get a PartialDMChannel?

Regardless, I think we should ensure this always returns a DMChannel by doing something like `return channel?.type === 'DM' && whateverCheckThatIsFalseForPartial`

Favna (changes requested):

packages/discord.js-utilities/src/lib/type-guards.ts
kyranet: > In what instance does a bot get a PartialDMChannel?

When you use the `CHANNELS` partial and you get a DM message.

Favna (approved):

Favna (approved):

PR #166, opened 2021-09-12
fix(snowflake): fixed snowflake generating duplicate IDs

### LINKED ISSUE #154: bug: Snowflake generating the same ID twice before incrementing
**Describe the bug**

image
image

Here's how I generated the IDs
```ts
   public static readonly snowflake = new Snowflake(new Date('2005-08-07T00:00:00.000Z'));
   /**
     * Generate a buttonID for discord
     * Looks like loop-q:a:n-2132321332112
     * @param command The name of the command to run
     * @param args The arguments to pass the command
     */
    public static generateButtonID(command: string, args?: string[]): string {
        const test = `${command}-${args ? args.join(':') : 'null'}-${this.snowflake.generate()}`;
        console.log(test);
        return test;
    };
```
**To Reproduce**

1. Running the Snowflake#generate() method I guess?

Favna:
fixes #154

### DIFF
--- a/packages/snowflake/tests/lib/Snowflake.test.ts
+++ b/packages/snowflake/tests/lib/Snowflake.test.ts
@@ -74,16 +74,16 @@ describe('Snowflake', () => {
 			const snowflake = new Snowflake(sampleEpoch);
 
 			const arrayOf10Snowflakes = [
-				snowflake.generate().toString(),
-				snowflake.generate().toString(),
-				snowflake.generate().toString(),
-				snowflake.generate().toString(),
-				snowflake.generate().toString(),
-				snowflake.generate().toString(),
-				snowflake.generate().toString(),
-				snowflake.generate().toString(),
-				snowflake.generate().toString(),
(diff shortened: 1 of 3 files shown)


packages/snowflake/tests/lib/Snowflake.test.ts
```diff
@@ -69,20 +69,42 @@ describe('Snowflake', () => {
...
+				snowflake.generate().toString(),
+				snowflake.generate().toString(),
+				snowflake.generate().toString(),
+				snowflake.generate().toString(),
+				snowflake.generate().toString(),
+				snowflake.generate().toString(),
+				snowflake.generate().toString(),
+				snowflake.generate().toString(),
+				snowflake.generate().toString(),
+				snowflake.generate().toString()
```
vladfrangu, line 86: You don't need toString

vladfrangu (approved):

vladfrangu (approved):

PR #217, opened 2021-11-14
fix(eslint-config): allow type and regular imports

Lioness100:
Use [`@typescript-eslint/no-duplicate-imports`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-duplicate-imports.md) instead of `no-duplicate-imports`.

> This rule extends the base `eslint/no-duplicate-imports` rule. This version adds support for type-only import and export.


favna (approved):
wait wtf I thought it was already this way,

PR #232, opened 2021-12-08
fix(snowflake): remove env-based defaults

kyranet:
This makes browser bundling easier, besides, process IDs and thread IDs are 32-bit integers, which we cannot fit in 5 bits, so there wasn't a real guarantee of uniqueness across threads or servers.

### DIFF
--- a/packages/snowflake/src/lib/Snowflake.ts
+++ b/packages/snowflake/src/lib/Snowflake.ts
@@ -1,11 +1,9 @@
 /* eslint-disable @typescript-eslint/explicit-member-accessibility */
 
-import { threadId } from 'node:worker_threads';
-
 // For testing purposes, we'll assign processId and workerId
 // to 1n so it's easier to match the generated snowflakes:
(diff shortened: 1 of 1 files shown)


packages/snowflake/src/lib/Snowflake.ts
```diff
@@ -1,11 +1,9 @@
 /* eslint-disable @typescript-eslint/explicit-member-accessibility */
 
-import { threadId } from 'node:worker_threads';
-
 // For testing purposes, we'll assign processId and workerId
```
vladfrangu, line 3: Drop the comments

vladfrangu (changes requested):

packages/snowflake/src/lib/Snowflake.ts
kyranet: Donette.

vladfrangu (approved):

favna (approved):

PR #254, opened 2022-01-10
feat(discord.js-utilities): add `PaginatedMessageEmbedFields`

noftaly:
`PaginatedMessageEmbedFields` (this new one) differs from `PaginatedFieldMessageEmbed` (Kanama's one) as the items here are whole fields, that are added to the embed, whereas `PaginatedFieldMessageEmbed` concatenates the items in a single field with a given formatter function.

I believe this one is quite useful too and will find its lot of users.

### DIFF
--- a/packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessageEmbedFields.ts
+++ b/packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessageEmbedFields.ts
@@ -1,0 +1,148 @@
+import { EmbedLimits } from '@sapphire/discord-utilities';
+import { isFunction } from '@sapphire/utilities';
+import { EmbedFieldData, MessageEmbed, MessageEmbedOptions } from 'discord.js';
+import { PaginatedMessage } from './PaginatedMessage';
+
+/**
+ * This is a utility of {@link PaginatedMessage}, except it exclusively paginates the fields of an embed.
+ * You must either use this class directly or extend it.
+ *
+ * It differs from PaginatedFieldMessageEmbed as the items here are whole fields, that are added to the embed,
+ * whereas PaginatedFieldMessageEmbed concatenates the items in a single field with a given formatter function.
+ *
+ * @example
+ * ```typescript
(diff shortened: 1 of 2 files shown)


packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessageEmbedFields.ts
```diff
@@ -0,0 +1,135 @@
...
+	 *
+	 * new PaginatedMessageEmbedFields()
+	 * 	.setTemplate({ title: 'Test pager embed', color: '#006080' })
+	 * 	.setItems([{ name: 'My field', value: 'The field\'s value' }])
+	 * 	.make()
+	 * 	.run(message);
+	 * ```
+	 */
+	public setTemplate(template: MessageEmbed | MessageEmbedOptions): this {
+		this.embedTemplate = template instanceof MessageEmbed ? template : new MessageEmbed(template);
```
favna, line 83:
```suggestion
		this.embedTemplate = this.resolveTemplate(template);
```

Also support an embed callback. 

```ts
private resolveTemplate(template: MessageEmbed | MessageEmbedOptions | ((embed: MessageEmbed) => MessageEmbed)) {
	if (template instanceof MessageEmbed) {
		return template;
	} else if (isFunction(template)) {
		return template(new MessageEmbed());
	}

	return new MessageEmbed(template);
}
```

This way it matches the possibilities for `addPageEmbed`. `isFunction` is from `@sapphire/utilities`
Also add an example for this of course.

favna (changes requested):

packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessageEmbedFields.ts
noftaly: Will do. Do i also add this to PaginatedFieldMessageEmbed.setTemplate in another commit?

favna: Might as well yeah actually

favna (approved):

PR #258, opened 2022-01-14
feat(decorators): add container to `ApplyOptions` callback

favna:
BREAKING CHANGE: When using `@ApplyOptions` with a callback function, the single parameter of `PieceContex` been changed to an object, which also has a property of `Container`. Migration change is `@ApplyOptions<Listener.Options>(({ context }) => ({`

### DIFF
--- a/packages/decorators/src/piece-decorators.ts
+++ b/packages/decorators/src/piece-decorators.ts
@@ -1,4 +1,5 @@
-import type { Piece, PieceContext, PieceOptions } from '@sapphire/framework';
+import { container, type Piece, type PieceContext, type PieceOptions } from '@sapphire/framework';
+import type { Container } from '@sapphire/pieces';
 import type { Ctor } from '@sapphire/utilities';
 import { createClassDecorator, createProxy } from './utils';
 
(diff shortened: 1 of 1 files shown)


kyranet:
What's the use for this change? Can't developers just import container from framework?

vladfrangu:
This very much is a semver change btw, even if <1% uses this style

favna:
> What's the use for this change? Can't developers just import container from framework?

If "could developers just do X" would be a valid reason not to do something then the following would also be valid:

What's the use of a framework? Can't developers just make their own?

So just like with framework and with anything we do the reason is very simple: making things easier.
Why do we export namespaces? To save on type only imports
Why do we have multiple subclasses for paginated messages? To have niche cases covered
etc.

vladfrangu:
I'd argue container should be after context, because more often than not you need.. the context, not the container.
Alternatively, make it an object

favna:
I'll make it an object... tomorrow-ish. Too tired now.

kyranet:
> What's the use of a framework? Can't developers just make their own?

Don't mix/confuse things, most users import `Command`, `Listener`, and other classes from `@sapphire/framework`, getting access to the container is as easy as adding it to the import list.

Same thing, different location.

Furthermore, the only usecase for this (that I saw in the support server) is so people could do `container.client.ws`, that's unnecessary, they can just do `'ws'`.

favna:
> Furthermore, the only usecase for this (that I saw in the support server) is so people could do `container.client.ws`, that's unnecessary, they can just do `'ws'`.

No that was me misunderstanding. They actually wanted to access `container.client.db.connection` which cannot be changed to `'db.connection'` or something.

kyranet:
> No that was me misunderstanding. They actually wanted to access `container.client.db.connection` which cannot be changed to `'db.connection'` or something.

That's easily fixable.

packages/decorators/src/piece-decorators.ts
```diff
@@ -49,14 +49,21 @@ import { createClassDecorator, createProxy } from './utils';
  * }
  * ```
  */
-export function ApplyOptions<T extends PieceOptions>(optionsOrFn: T | ((container: Container, context: PieceContext) => T)): ClassDecorator {
+export function ApplyOptions<T extends PieceOptions>(
+	optionsOrFn: T | (({ container, context }: ApplyOptionsCallbackParameters) => T)
```
vladfrangu, line 53: Why are you destructuring these? How is this even valid o.o

vladfrangu (changes requested):

packages/decorators/src/piece-decorators.ts
favna: Hi I'm silly. How are you?

vladfrangu (approved):

PR #259, opened 2022-01-15
feat(PaginatedMessage): made it possible to add custom `link` buttons with `setActions` / `addActions` / `addAction`

favna:
This makes it possible to add `LINK` style buttons to the paginated message's actions.
Previously it was set up to lock out this possibility.
I probably had some issues implementing it in a sensible way and gave up, I don't remember.

I also made it possible to set a second optional parameter of `setActions` to true
to have the library merge in `PaginatedMessage.defaultActions`, something that is a
common use case when using this particular function.
This change will save users from having to manually add the default actions.

I'm not sure if this change should be considered `semver:major`. @vladfrangu what do you say?
I'm primarily asking because `PaginatedMessageAction` was changed from an `interface` to a `type`
which subsequently caused this issue at least in Skyra's code:

Easily fixed of course by just changing it to a `type`, but the code did not compile.
Is that `semver:major`?

```typescript
type HelpPaginatedMessageAction = PaginatedMessageAction & {
	selectMenuIndex?: 'set-1' | 'set-2';
};
```

### DIFF
--- a/packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
+++ b/packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
@@ -28,7 +28,7 @@ import type {
 	PaginatedMessageSelectMenuOptionsFunction,
 	PaginatedMessageWrongUserInteractionReplyFunction
 } from './PaginatedMessageTypes';
-import { createPartitionedMessageRow, isMessageButtonInteraction } from './utils';
+import { actionIsButtonOrMenu, createPartitionedMessageRow, isMessageButtonInteraction } from './utils';
 
 /**
  * This is a {@link PaginatedMessage}, a utility to paginate messages (usually embeds).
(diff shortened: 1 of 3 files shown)


packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
```diff
@@ -816,25 +886,28 @@ export class PaginatedMessage {
...
+				await action.run({
+					interaction,
+					handler: this,
+					author: targetUser,
+					channel,
+					response: this.response!,
+					collector: this.collector!
+				});
+
+				const newIndex = previousIndex === this.index ? previousIndex : this.index;
```
vladfrangu, line 902: Doesn't this end up as this.index in both cases?

vladfrangu (approved):

packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
favna: uuhhhh.. that's a good point.

PR #263, opened 2022-01-17
fix(prettier-config): typo

FC5570:


ItsMrCube (changes requested):
I would write “Communities” instead of “Community”

favna:
There were a few others as well so I fixed them all in 1 go

PR #266, opened 2022-01-20
docs(readme): update npm badge

Tokipudi:
Fixed a wrong label for an NPM link.


favna (approved):
I low-key wondered when we'd see the first PR from you with how active you've been in the server.

PR #274, opened 2022-01-27
feat(package): add @sapphire/result package

### LINKED ISSUE #273: request: port `Result` code from framework to new package `@sapphire/result`

### Feature description

More and more do I see use cases for https://github.com/sapphiredev/framework/blob/main/src/lib/parsers/Result.ts in other places that do not depend on `@sapphire/framework` so it would be nice if this code was in a new package called`@sapphire/result` instead.

### Desired solution

First of all, the code from the https://github.com/sapphiredev/framework/blob/main/src/lib/parsers/Result.ts file to be ported to `@sapphire/result`. Then, once an update has been released, a follow up PR to be made in framework to remove the code there and instead re-export it from utilities.

Secondly, as we would like @sapphire/result to work on the web, be small and standalone, and not depend on any external packages, we should include the types `Ok` and `Err` types from Lexure directory into our own code with a proper license header copy in 1:1 into utilities.

### Alternatives considered

**Including the code in `@sapphire/utilities`**

We could copy the code into `@sapphire/utilities`, but that unnecessarily increases the bundle size of any package that requires only `Result` such as @sapphire/shapeshift.

### Additional context

_No response_

c43721:
Fixes #273 

Copies Result code from both framework and Lexure and exposes both `Result` and `Maybe` into it's own package. See issue for more details.


favna:
oh ffs

packages/result/src/lib/Parsers.ts
```diff
@@ -0,0 +1,34 @@
...
+export function from<T, E = unknown>(cb: (...args: unknown[]) => T): Result<T, E> {
+	try {
+		return ok(cb());
+	} catch (error) {
+		return err(error as E);
+	}
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+function isFunction(input: unknown): input is Function {
```
c43721, line 19: Typescript isn't too happy without the strong Function type from this. Any way around this?

favna: Nah there isn't. We have it this way in utilities as well for that reason.

favna, line 19: `function` expressions get lifted up by JS so the order of definition doesn't matter. You can place this one at the bottom so the exports are above it.

packages/result/src/lib/Parsers.ts
```diff
@@ -0,0 +1,34 @@
+import { err, ok, Result } from './Result';
+
+type Awaitable<T> = PromiseLike<T> | T;
```
favna, line 3: Also put this type below

So the structure is

- from
- fromAsync
- isFunction
- Awaitable

favna (changes requested):

favna (approved):

favna (dismissed):

favna (approved):

PR #275, opened 2022-01-27
fix: fixed date cloning

### LINKED ISSUE #272: bug: `Date` parsing in `deepClone`
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the bug

There is a critical bug in `deepClone` in `@sapphire/utilities` (which then also affects `mergeDefault`) where it is incapable of parsing dates. Instead of duplicating the Date, it assigns the current date.

### Steps To Reproduce

```js
const { deepClone } = require('@sapphire/utilities');

const originalObject = { dateTime: new Date('2020-01-01T12:45:00.000Z') };

const clonedObject = deepClone(originalObject);

console.log(obj);
// { dateTime: 2020-01-01T12:45:00.000Z }

console.log(clonedObject);
// { dateTime: 2022-01-26T23:31:03.734Z }
```

Not the same date!

### Expected behavior

The date has to be the same

### Screenshots

image

### Additional context

I discovered this when we ran into the bug with code at work

favna:
fixes #272

wew lad

### DIFF
--- a/packages/utilities/src/lib/deepClone.ts
+++ b/packages/utilities/src/lib/deepClone.ts
@@ -11,7 +11,7 @@ export function deepClone<T>(source: T): T {
 	}
 
 	if (source instanceof Date) {
-		const output = new (source.constructor as DateConstructor)(source.toISOString());
+		const output = new (source.constructor as DateConstructor)(source);
 
 		return output as unknown as T;
 	}
--- a/packages/utilities/src/lib/deepClone.ts
+++ b/packages/utilities/src/lib/deepClone.ts
@@ -10,6 +10,12 @@ export function deepClone<T>(source: T): T {
 		return source;
 	}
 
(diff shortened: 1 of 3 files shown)


packages/utilities/src/lib/deepClone.ts
```diff
@@ -10,6 +10,12 @@ export function deepClone<T>(source: T): T {
 		return source;
 	}
 
+	if (source instanceof Date) {
+		const output = new (source.constructor as DateConstructor)(source.toISOString());
```
kyranet, line 14:
```suggestion
		const output = new (source.constructor as DateConstructor)(source);
```

kyranet (changes requested):

kyranet (approved):

vladfrangu (approved):

c43721 (approved):

PR #276, opened 2022-01-28
fix(PaginatedMessageEmbedFields): custom template with fields

kaname-png:


noftaly:
This was on purpose, but if you feel like it is needed, just add a check to be sure not to have more than 25 fields :)

noftaly (approved):

c43721 (approved):

favna (approved):

PR #283, opened 2022-02-02
fix: fixed issues with `PaginatedMessage`

favna:
Well, that was mega awkward but apparently, the way we created the handler it was triggering on *all* handlers in the current channel as opposed to just the one that actually triggered it.

### DIFF
--- a/packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
+++ b/packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
@@ -1,5 +1,6 @@
 import { Time } from '@sapphire/time-utilities';
 import { deepClone, isFunction, isNullish, isObject } from '@sapphire/utilities';
+import type { APIMessage } from 'discord-api-types/v9';
 import {
 	Constants,
 	Formatters,
@@ -20,7 +21,7 @@ import {
 	type WebhookEditMessageOptions
 } from 'discord.js';
 import { MessageBuilder } from '../builders/MessageBuilder';
-import { isGuildBasedChannel } from '../type-guards';
+import { isGuildBasedChannel, isMessageInstance } from '../type-guards';
 import type {
 	PaginatedMessageAction,
 	PaginatedMessageEmbedResolvable,
@@ -115,7 +116,7 @@ export class PaginatedMessage {
 	/**
 	 * The response message used to edit on page changes.
 	 */
-	public response: Message | CommandInteraction | SelectMenuInteraction | ButtonInteraction | null = null;
+	public response: APIMessage | Message | CommandInteraction | SelectMenuInteraction | ButtonInteraction | null = null;
 
 	/**
 	 * The collector used for handling component interactions.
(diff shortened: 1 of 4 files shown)


packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
```diff
@@ -867,16 +868,14 @@ export class PaginatedMessage {
...
 		} else if (runsOnInteraction(messageOrInteraction)) {
-			this.response = messageOrInteraction;
-
-			if (this.response.replied || this.response.deferred) {
-				await this.response.editReply(page as MessageOptions);
+			if (messageOrInteraction.replied || messageOrInteraction.deferred) {
+				this.response = await messageOrInteraction.editReply(page);
 			} else {
-				await this.response.reply(page as MessageOptions);
+				this.response = await messageOrInteraction.reply({ fetchReply: true, ...page });
```
vladfrangu, line 878: If we provide fetchReply to false in the lage it won't fetch, if we provide ephemeral in the page it won't work

packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
```diff
@@ -891,11 +890,25 @@ export class PaginatedMessage {
 	protected setUpCollector(channel: TextBasedChannel, targetUser: User): void {
 		if (this.pages.length > 1) {
 			this.collector = new InteractionCollector<MessageComponentInteraction>(targetUser.client, {
-				filter: (interaction) => interaction.isMessageComponent() && this.actions.has(interaction.customId),
+				filter: (interaction) =>
+					!isNullish(this.response) &&
+					interaction.message.id === this.response?.id &&
```
favna, line 895: Theoretically, these lines are no longer needed because of the message filter below but I don't feel confident enough to remove them. Thoughts Vladdy/Kyra?

I did a quick test with them disabled and it does seem to work fine still.

packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
favna: As far as TS is concerned it's not possible for a page to be `ephemeral`. `page` is of type `MessageOptions | WebhookEditMessageOptions`, which are https://github.com/discordjs/discord.js/blob/stable/typings/index.d.ts#L5144-L5155 and https://github.com/discordjs/discord.js/blob/stable/typings/index.d.ts#L5673-L5676

As for the `fetchReply`, that too is not possible

Anyway for sanity checks I made it `{ ...page, fetchReply: true, ephemeral: false }`. This does still work for pre-existing ephemeral responses because those would use `editReply` (i.e. if I first defer then call `.run`)

packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
vladfrangu: I have no idea..

vladfrangu (approved):

PR #291, opened 2022-02-12
feat(PaginatedMessage): emit warning when running in a DM channel without required client options

favna:
This introduces a new potential emitted warning when the client is not
properly configured to handle DM commands.

Whether the warning is emitted at all can be configured so people can choose to not get it at all.
It will also only be emitted once per instance of PaginatedMessage.

On the note of the warning being per instance of PaginatedMessage, yes that does often mean that
people will get it for every time that `run` is called because it's common (and recommended)
to call `new PaginatedMessage` inside of the `*Run` method.
I don't think it's fitting to keep track of that boolean outside of the class though.

Calling run twice example:

### DIFF
--- a/packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
+++ b/packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
@@ -808,7 +808,7 @@ export class PaginatedMessage {
 			) {
 				process.emitWarning(
 					[
-						'PaginatedMessage initiated in a DM channel without the client having required partial configured.',
+						'PaginatedMessage was initiated in a DM channel without the client having the required partial configured.',
 						'If you want PaginatedMessage to work in DM channels then make sure you start your client with "CHANNEL" added to "client.options.partials".',
 						'If you do not want to be alerted about this in the future then you can disable this warning by setting "PaginatedMessage.emitPartialDMChannelWarning" to "false", or use "setEmitPartialDMChannelWarning(false)" before calling "run".'
 					].join('\n'),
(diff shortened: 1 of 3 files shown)


packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
```diff
@@ -753,46 +799,67 @@ export class PaginatedMessage {
...
+		// If there is no channel then exit early and potentially emit a warning
+		if (!messageOrInteraction.channel) {
+			if (
+				this.emitPartialDMChannelWarning &&
+				!this.hasEmittedPartialDMChannelWarning &&
+				!messageOrInteraction.client.options.partials?.includes('CHANNEL')
+			) {
+				process.emitWarning(
+					[
+						'PaginatedMessage initiated in a DM channel without the client having required partial configured.',
```
vladfrangu, line 811:
```suggestion
						'PaginatedMessage was initiated in a DM channel without the client having the required partial configured.',
```

vladfrangu (changes requested):

packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
```diff
@@ -753,46 +807,106 @@ export class PaginatedMessage {
...
 			}
 
-			await this.resolvePagesOnRun();
+			await safelyReplyToInteraction({
+				messageOrInteraction,
+				interactionEditReplyContent: this.#thisMazeWasNotMeantForYouContent,
+				interactionReplyContent: { ...this.#thisMazeWasNotMeantForYouContent, ephemeral: true },
+				componentUpdateContent: this.#thisMazeWasNotMeantForYouContent,
+				messageMethod: 'reply',
+				messageMethodContent: this.#thisMazeWasNotMeantForYouContent
```
vladfrangu, line 867: Thats a lot of duplications... 😓

vladfrangu (approved):

packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts
favna: Better than duplicating the whole method thrice :(

PR #302, opened 2022-02-26
feat(discord-utilities): Add token regex

Commandtechno:
A token regex can be useful for automatically resetting tokens, and validating tokens before making requests or connecting to the gateway.

Source is from searching TOKEN_REGEX in https://canary.discord.com/assets/2fa49c2711296797d83c.js

image

### DIFF
--- a/packages/discord-utilities/src/lib/regexes.ts
+++ b/packages/discord-utilities/src/lib/regexes.ts
@@ -120,7 +120,7 @@ export const SnowflakeRegex = /^(?<id>\d{17,19})$/;
  * You can use the name of the capture group to determine if the validated token was configured for a user with Multi-Factor Authentication, for a user without Multi-Factor Authentication, or for a bot application.
  * If both capture groups are undefined, then the token is invalid.
  */
-export const TokenRegex = /(?<mfaToken>mfa\.[a-z0-9_-]{20,})|(?<botToken>[a-z0-9_-]{23,28}\.[a-z0-9_-]{6,7}\.[a-z0-9_-]{27})/i;
+export const TokenRegex = /(?<mfaToken>mfa\.[a-z0-9_-]{20,})|(?<basicToken>[a-z0-9_-]{23,28}\.[a-z0-9_-]{6,7}\.[a-z0-9_-]{27})/i;
 
 /**
  * Regex that can capture a Twemoji (Twitter Emoji)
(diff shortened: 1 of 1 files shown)


vladfrangu (dismissed):

packages/discord-utilities/src/lib/regexes.ts
```diff
@@ -110,6 +110,12 @@ export const RoleMentionRegex = /^<@&(?<id>\d{17,19})>$/;
  */
 export const SnowflakeRegex = /^(?<id>\d{17,19})$/;
 
+/**
+ * Regex that can capture a Discord Token
+ * @raw /(mfa\.[a-z0-9_-]{20,})|([a-z0-9_-]{23,28}\.[a-z0-9_-]{6,7}\.[a-z0-9_-]{27})/i
+ */
+export const TokenRegex = /(mfa\.[a-z0-9_-]{20,})|([a-z0-9_-]{23,28}\.[a-z0-9_-]{6,7}\.[a-z0-9_-]{27})/i;
```
favna, line 117:
```suggestion
/**
 * Regex that can capture a Discord Token
 * @raw `/(?<mfaToken>mfa\.[a-z0-9_-]{20,})|(?<sfaToken>[a-z0-9_-]{23,28}\.[a-z0-9_-]{6,7}\.[a-z0-9_-]{27})/i`
 * @remark Capture group 1 can be used to retrieve a token for a User that has Multi-Factor Authentication enabled. It is named `mfaToken`.
 * @remark Capture group 2 can be used to retrieve a token for a Bot application, or a User that does not have Multi-Factor Authentication enabled. It is named `sfaToken`.
 * @remark For a valid token, either Capture group 1 or Capture group 2 will always be undefined, whereas the other group will then be defined and contain the matched token.
 * You can use the name of the capture group to determine if the validated token was configured for a user with Multi-Factor Authentication.
 * If both capture groups are undefined, then the token is invalid.
 */
export const TokenRegex = /(?<mfaToken>mfa\.[a-z0-9_-]{20,})|(?<sfaToken>[a-z0-9_-]{23,28}\.[a-z0-9_-]{6,7}\.[a-z0-9_-]{27})/i;
```

favna (changes requested):

packages/discord-utilities/src/lib/regexes.ts
vladfrangu: sure

vladfrangu (approved):

Commandtechno:
I changed botToken to basicToken since this regex also matches tokens for users without 2FA enabled (tested on my alt's token)

favna (approved):

kyranet (approved):

PR #318, opened 2022-03-21
feat: add argument parser

imranbarbhuiya:
copy-pasted from https://github.com/sapphiredev/framework to make argument parser available outside the framework

before starting I thought I can replace the command part with a simple object containing the `run` method but now I realized why vladdy said it's semi-independent xD. For now, I've added a simple structure for command but ik it's not acceptable so pls lmk how should I  replace the command part.


packages/discord-argument/CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+# Change Log
```
favna, line 1: You can remove this file. Lerna will generate it.

packages/discord-argument/README.md
```diff
@@ -0,0 +1,49 @@
+<div align="center">
+
+![Sapphire Logo](https://cdn.skyra.pw/gh-assets/sapphire-banner.png)
+
+# @sapphire/discord-argument
+
+**Template for Sapphire Repositories.**
```
favna, line 7: Template

packages/discord-argument/README.md
```diff
@@ -0,0 +1,49 @@
...
+</div>
+
+## Features
+
+-   This
+-   Needs
+-   Some
+-   More
+-   Filling
+-   In
```
favna, line 22: TODO

packages/discord-argument/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/docs/en/emoji-key)):
+
+<!-- ALL-CONTRIBUTORS-LIST:START - Do not remove or modify this section -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-disable -->
+<!-- markdownlint-enable -->
+<!-- prettier-ignore-end -->
+
+<!-- ALL-CONTRIBUTORS-LIST:END -->
```
favna, line 47: Run `generate` from the all-contributors cli: https://github.com/all-contributors/cli

packages/discord-argument/jest.config.ts
```diff
@@ -0,0 +1,16 @@
+import type { Config } from '@jest/types';
```
favna, line 1: make this mjs

packages/discord-argument/package.json
```diff
@@ -0,0 +1,60 @@
+{
+	"name": "@sapphire/discord-argument",
+	"version": "0.0.1",
```
favna, line 3:
```suggestion
	"version": "0.0.0",
```

packages/discord-argument/package.json
```diff
@@ -0,0 +1,60 @@
...
+	"description": "template",
```
favna, line 4: template

packages/discord-argument/package.json
```diff
@@ -0,0 +1,60 @@
...
+		"discord.js"
+	],
+	"bugs": {
+		"url": "https://github.com/sapphiredev/utilities/issues"
+	},
+	"publishConfig": {
+		"access": "public"
+	},
+	"dependencies": {
+		"@sapphire/discord.js-utilities": "workspace:^",
```
favna, line 54:
```suggestion
		"@sapphire/discord.js-utilities": "*",
```

packages/discord-argument/package.json
```diff
@@ -0,0 +1,60 @@
...
+		"@sapphire/pieces": "^3.3.0",
+		"@sapphire/result": "workspace:^",
```
favna, line 56:
```suggestion
		"@sapphire/result": "*",
```

packages/discord-argument/package.json
```diff
@@ -0,0 +1,60 @@
...
+	"keywords": [
+		"@sapphire/argument",
+		"bot",
+		"typescript",
+		"ts",
+		"yarn",
+		"discord",
+		"sapphire",
+		"standalone",
+		"discord.js"
```
favna, line 45:
```suggestion
		"standalone",
		"discord.js",
		"argument",
		"parser"
```

packages/discord-argument/package.json
```diff
@@ -0,0 +1,60 @@
...
+	"homepage": "https://github.com/sapphiredev/utilities/tree/main/packages/discord-argument",
+	"scripts": {
+		"lint": "eslint src --ext ts --fix -c ../../.eslintrc",
+		"build": "tsup && tsc -b src",
+		"prepublishOnly": "rollup-type-bundler -e @sapphire/result @sapphire/pieces @sapphire/discord.js-utilities discord.js"
+	},
+	"repository": {
+		"type": "git",
+		"url": "git+https://github.com/sapphiredev/utilities.git",
+		"directory": "packages/argument"
```
favna, line 26:
```suggestion
		"directory": "packages/argument-parser"
```

packages/discord-argument/package.json
```diff
@@ -0,0 +1,60 @@
+{
+	"name": "@sapphire/discord-argument",
```
favna, line 2:
```suggestion
	"name": "@sapphire/argument-parser",
```

Name the package like this. Reflect this everywhere (readme, folder name, etc)

packages/discord-argument/src/lib/structures/Command.ts
```diff
@@ -0,0 +1,66 @@
+import type { Message } from 'discord.js';
```
favna, line 1: Idk about this tbh. Pretty much what Vladdy said. It feels wrong to have that here. Technically not an issue to framework because it can extend and then export but ... I just don't know. I want to hear @vladfrangu  and @kyranet's opinions here

packages/discord-argument/src/lib/structures/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './Argument';
```
favna, line 1: remove this index.ts and just import with an extra path jump elsewhere. That's more future proof to avoid circular references.

favna (changes requested):
Also needs updating
- [.npm-deprecaterc.yml](https://github.com/sapphiredev/utilities/blob/main/.npm-deprecaterc.yml)
- [.all-contributorsrc](https://github.com/sapphiredev/utilities/blob/main/.all-contributorsrc)
- [README.md](https://github.com/sapphiredev/utilities/blob/main/README.md) (badge)
- [lerna.json](https://github.com/sapphiredev/utilities/blob/main/lerna.json) (`ignoreChanges`)

packages/discord-argument/src/lib/structures/Command.ts
vladfrangu: I don't like this at all. If we really want to abstract args out of the framework, it should provide a way to register resolvers by name (`argsResolvers.register('name', IArgument)`) and use that instead. In no way are we re-extending in framework from a module

packages/argument-parser/src/lib/errors/Identifiers.ts
```diff
@@ -0,0 +1,43 @@
+export const enum Identifiers {
```
vladfrangu, line 1:
```suggestion
export enum Identifiers {
```

packages/argument-parser/src/lib/parsers/Args.ts
```diff
@@ -0,0 +1,750 @@
...
+import type { URL } from 'url';
+import { ArgumentError } from '../errors/ArgumentError';
+import { Identifiers } from '../errors/Identifiers';
+import { UserError } from '../errors/UserError';
+import type { EmojiObject } from '../resolvers/emoji';
+import type { Argument, IArgument } from '../structures/Argument';
+import { ArgumentStore } from '../structures/ArgumentStore';
+import type { Command } from '../structures/Command';
+
+container.stores.register(new ArgumentStore().registerPath(join(__dirname, '..', 'arguments')));
```
vladfrangu, line 28: Does stores actually exist outside framework? And even if so, nothing will load them 🤔 

I'd expect more of an ArgsStore.register('name', (parameter, whatever) => {}) (`register<T>(name: string, arg: IArgument<T>)`)

packages/argument-parser/src/lib/parsers/Args.ts
```diff
@@ -0,0 +1,750 @@
...
+
+	/**
+	 * The command that is being run.
+	 */
+	public readonly command: Command;
+
+	/**
+	 * The context of the command being run.
+	 */
+	public readonly commandContext: Command.RunContext;
```
vladfrangu, line 47: None of these should be present here, at best .run should get an extensible context which might have anything or nothing

packages/argument-parser/src/lib/structures/Command.ts
```diff
@@ -0,0 +1,66 @@
+import type { Message } from 'discord.js';
+import * as Lexure from 'lexure';
+import { Args } from '../parsers/Args';
+import { FlagStrategyOptions, FlagUnorderedStrategy } from '../utils/strategies/FlagUnorderedStrategy';
+
+export namespace Command {
+	export type RunContext = Record<PropertyKey, unknown>;
+}
+
+export abstract class Command {
```
vladfrangu, line 10: This should go

vladfrangu (changes requested):

packages/argument-parser/src/lib/parsers/Args.ts
imranbarbhuiya: I never used `pieces` so I don't know how this works. But I only checked the lib to know how can I use it (just to use it here 😄 . I'll go with the 2nd option.

imranbarbhuiya:
After a long exam break, I forgot what how I was planning to fix this so I'm closing it now. I'll reopen it If I get a better idea to allow registering commands.

PR #329, opened 2022-04-17
chore: setup git-cliff

favna:
**TODO**

- [x] Figure out a way to bump package versions easily.
   - `standard-version` doesn't consider the current directory only, you can see in https://github.com/sapphiredev/utilities/actions/runs/2180006374 that it tried doing major versions for everything when in reality none should've been: https://github.com/sapphiredev/utilities/runs/6052668420?check_suite_focus=true
   - `yarn version` requires manually finding out if it needs to be Patch, Minor or Major
   - Other??
- [x] Validate the above setup with the continuous delivery script
- [x] Ensure continuous delivery script will run normally again instead of just logging
- [x] Remove copy script again
- [x] Make package out of bump script


.github/workflows/continuous-delivery.yml
```diff
@@ -39,15 +39,46 @@ jobs:
...
+        working-directory: packages/
         run: |
           # Resolve the tag to be used. "next" for push events, "pr-{prNumber}" for dispatch events.
           TAG=$([[ ${{ github.event_name }} == 'push' ]] && echo 'next' || echo 'pr-${{ github.event.inputs.prNumber }}')
-          BRANCH=$(echo '${{ github.event.inputs.ref || 'main' }}')
 
-          # Bump and publish
-          yarn lerna publish --yes --conventional-prerelease --ignore-scripts --no-verify-access --no-git-tag-version --no-push --pre-dist-tag ${TAG} --preid "${TAG}.$(git rev-parse --verify --short HEAD)" --allow-branch ${BRANCH}
+          yarn config set npmAuthToken ${NODE_AUTH_TOKEN}
+          yarn config set npmPublishRegistry "https://registry.yarnpkg.com"
```
vladfrangu, line 53: Are these two lines actually needed?

vladfrangu (approved):

.github/workflows/continuous-delivery.yml
favna: Yes

PR #343, opened 2022-05-01
feat: add new eslint rules

imranbarbhuiya:


favna (commented):
Sorry but this will be rejected
- `consistent-type-imports` doesn't product good looking code. For example from:
```ts
import { Message as DJSMessage, PermissionResolvable, Permissions, PermissionString } from 'discord.js';
```
   it makes:
```ts
import type { Message as DJSMessage, PermissionResolvable, PermissionString } from 'discord.js';
import { Permissions } from 'discord.js';
```
   as opposed to what we would actually want:
```ts
import { Permissions, type Message as DJSMessage, type PermissionResolvable, type PermissionString } from 'discord.js';
```   
- `no-unnecessary-condition` already messes with the code base of this package, let alone others. And all of the ones that it errors on are false positives.
- `no-unnecessary-type-assertion` has historically given us a lot of trouble when it would incorrectly assert something to be type A when it really wasn't.

PR #344, opened 2022-05-04
chore: bump jest

imranbarbhuiya:


favna (changes requested):
Jsut 1 thingy, you can remove the additional report from all configs that reside in `packages/`, CI uses the on the root. The subconfigs are primarily there to focus tests easier because Jests' CI flags aren't everything.

favna (approved):

PR #349, opened 2022-05-13
feat(discord.js-utilities): isGuildMember type guard

kaname-png:


packages/discord.js-utilities/src/lib/type-guards.ts
```diff
@@ -187,3 +188,12 @@ export function isNsfwChannel(channel: ChannelTypes | Nullish): boolean {
 export function isMessageInstance(message: APIMessage | Message): message is Message {
 	return message instanceof Message;
 }
+
+/**
+ * Checks whether a given member is an instance of {@link GuildMember}, and not {@link APIInteractionGuildMember} or {@link APIGuildMember}
```
favna, line 193:
```suggestion
 * Checks whether a given member is an instance of {@link GuildMember}, and not {@link APIInteractionGuildMember}, {@link APIGuildMember}, or {@link Nullish}
```

packages/discord.js-utilities/src/lib/type-guards.ts
```diff
@@ -187,3 +188,12 @@ export function isNsfwChannel(channel: ChannelTypes | Nullish): boolean {
...
+ * @param member The message to check
```
favna, line 194:
```suggestion
 * @param member The member to check
```

favna (changes requested):

favna (approved):

PR #353, opened 2022-05-15
fix: simplify tsup config

imranbarbhuiya:


favna (changes requested):
This will break a config such as:
```ts
createTsupConfig()
```
And it will only work for:
```ts
createTsupConfig({ });
```
or, and similar to:
```ts
createTsupConfig({ format: ['esm'] });
```

imranbarbhuiya:
> This will break a config such as:
> 
> ```ts
> createTsupConfig()
> ```
> 
> And it will only work for:
> 
> ```ts
> createTsupConfig({ });
> ```
> 
> or, and similar to:
> 
> ```ts
> createTsupConfig({ format: ['esm'] });
> ```

No it won't I tested it.
image

favna (approved):

PR #354, opened 2022-05-21
fix(discord-utilities): update invite link regex

EvolutionX-10:


vladfrangu:
Please show strings that didn't match before but now do... Or just..what does this exactly change? It looks almost the same to me

EvolutionX-10:
alr @kyranet @vladfrangu


## Previous regex
image
image


## New regex
image
image

EvolutionX-10:
I dont think I'am able to see the image @imranbarbhuiya 🤔

imranbarbhuiya:
> I dont think I'am able to see the image @imranbarbhuiya 🤔

It's matching invalid codes like accepts space

EvolutionX-10:
How about this one @imranbarbhuiya 
`/(https:\/\/)?(www\.)?(((discord(app)?)?\.com\/invite)|((discord(app)?)?\.gg))\/(?<invite>\w+)/i`

packages/discord-utilities/src/lib/regexes.ts
```diff
@@ -29,10 +29,10 @@ export const DiscordHostnameRegex =
 
 /**
  * Regex that can can capture the code of Discord invite links
- * @raw `/^(?:https?:\/\/)?(?:www\.)?(?:discord\.gg\/|discord(?:app)?\.com\/invite\/)?(?<code>[\w\d-]{2,})$/i`
+ * @raw `/(https:\/\/)?(www\.)?(((discord(app)?)?\.com\/invite)|((discord(app)?)?\.gg))\/(?<invite>.+)/i`
```
imranbarbhuiya, line 32:
```suggestion
 * @raw `/discord(?:(?:app)?\.com\/invite|\.gg(?:\/invite)?)\/([\w-]{2,255})/i`
```

Maybe we can use this one which is used in djs

imranbarbhuiya (changes requested):

imranbarbhuiya:
> How about this one @imranbarbhuiya `/(https:\/\/)?(www\.)?(((discord(app)?)?\.com\/invite)|((discord(app)?)?\.gg))\/(?<invite>\S+)/i`

You are removing support for 
1. links starting with `http`. 
2. Previously only invite code was present in capturing block. Now u r capturing all the blocks.

EvolutionX-10:
> 1. You are removing support for links starting with `http`.

Nope not really
image

> 2. Previously only invite code was present in capturing block. Now u r capturing all the blocks.

Yeah i didn't notice that before, thanks, let @vladfrangu confirm for the d.js one, once he tells I'll commit the d.js one

imranbarbhuiya:
> > 1. You are removing support for links starting with `http`.
> 
> Nope not really image
> 

Well you missed `^` so it'll match
`adiscord.gg/abc`

EvolutionX-10:
I can imagine the abuse of `abcjoinmyserverdiscord.gg/sapphiredev` but anyways thats user's way to handle it

imranbarbhuiya:
> I can imagine the abuse of `abcjoinmyserverdiscord.gg/sapphiredev` but anyways thats user's way to handle it

still both 1 & 2 aren't resolved

EvolutionX-10:
what is not resolved?

imranbarbhuiya:
1. http
2. capture block

EvolutionX-10:
i don't get the capture group question

EvolutionX-10:
prettier f, gimme a min to fix

vladfrangu:
Is this regex the one in d.js slightly altered? Does it work properly now? Does it support http(s)-less invites? 👀

imranbarbhuiya:
> Is this regex the one in d.js slightly altered?
No
> Does it support http(s)-less invites?
yes

EvolutionX-10:
@vladfrangu now the regex is same as discord.js except there's a named group `code` now that captures the invite link code

packages/discord-utilities/src/lib/regexes.ts
```diff
@@ -29,10 +29,10 @@ export const DiscordHostnameRegex =
 
 /**
  * Regex that can can capture the code of Discord invite links
- * @raw `/^(?:https?:\/\/)?(?:www\.)?(?:discord\.gg\/|discord(?:app)?\.com\/invite\/)?(?<code>[\w\d-]{2,})$/i`
+ * @raw `/discord(?:(?:app)?\.com\/invite|\.gg(?:\/invite)?)\/(?<code>[\w-]{2,255})/gi`
  * @remark Capture group 1 is the invite URL's unique code. It is named `code`.
  */
-export const DiscordInviteLinkRegex = /^(?:https?:\/\/)?(?:www\.)?(?:discord\.gg\/|discord(?:app)?\.com\/invite\/)?(?<code>[\w\d-]{2,})$/i;
+export const DiscordInviteLinkRegex = /discord(?:(?:app)?\.com\/invite|\.gg(?:\/invite)?)\/(?<code>[\w-]{2,255})/gi;
```
vladfrangu, line 35: We probably want to add `(?:^|\b)` at the start and `(?:$|\b)` at the end

imranbarbhuiya: Check my message in discord. You can do like `"discord.gg/invite"`

EvolutionX-10: ### Looks like @vladfrangu's way works good @imranbarbhuiya 
image

vladfrangu (approved):

favna (approved):

PR #371, opened 2022-06-01
Fix tests and update dev deps

favna:
- test: fix fetch tests for Node 18
- chore: make husky hooks executable
- chore: bump dev deps


packages/fetch/tests/jest.setup.js
```diff
@@ -0,0 +1,11 @@
+let originalGlobalFetch = undefined;
+
+originalGlobalFetch = globalThis.fetch;
```
vladfrangu, line 3: Merge into one line

vladfrangu (changes requested):

packages/fetch/tests/jest.setup.js
favna: woops

PR #393, opened 2022-07-13
feat: lexure tests

UndiedGamer:


packages/lexure/tests/lexer.test.ts
```diff
@@ -0,0 +1,54 @@
+import { Lexer, QuotedParameter, WordParameter } from '../src';
+
+describe('The lexer tests', () => {
+	it('GIVEN input with no quotes THEN return the string as is', () => {
```
kyranet, line 4: We don't use `it()`, we use `test()`, apply this everywhere else as well:
```suggestion
	test('GIVEN input with no quotes THEN return the string as is', () => {
```

packages/lexure/tests/parser.test.ts
```diff
@@ -0,0 +1,50 @@
+import { EmptyStrategy, Lexer, Parser, ParserResult, PrefixedStrategy, WordParameter } from '../src';
+
+describe('Parser tests', () => {
```
kyranet, line 3:
```suggestion
describe('Parser', () => {
```

packages/lexure/tests/lexer.test.ts
```diff
@@ -0,0 +1,54 @@
...
+		const input = 'text without quotes';
+		const output = [...new Lexer().run(input)];
```
kyranet, line 6: Please don't use `output`, use `result`, and separate the initialization from the assertions. Apply this everywhere else as well.

kyranet (changes requested):
Missing tests for custom separators, as well as duplicated ones (think of double spaces).

Also, those two methods are untested:

https://github.com/sapphiredev/utilities/blob/dd0715e1bda4013012a929bf7b66d4e543d7c28d/packages/lexure/src/lib/lexer/streams/parameters/BaseParameter.ts#L8-L12

PR #388, opened 2022-07-02
feat: add `@sapphire/lexure` package

kyranet:
## TODO

- [ ] Docs
- [x] Tests

PR #394, opened 2022-07-13
refactor: change `const enum` for regular `enum`

favna:
To be frank I don't know why exactly but @vladfrangu told Kyra to remove it for lexure here https://github.com/sapphiredev/utilities/pull/388/files#r917235502 so I assume the others should go also


vladfrangu (approved):

kyranet:
Don't we have a preserve const enums thing?

iirc the only place where this matters, is in Deno and any project that uses isolated mode for some reason.

vladfrangu:
You...know very well the downsides of using const enums - including how much the TypeScript team recommends not using them

favna:
> You...know very well the downsides of using const enums - including how much the TypeScript team recommends not using them

For future reading this is reference to https://github.com/Microsoft/TypeScript/issues/30590 and similar issues

PR #404, opened 2022-07-28
feat(Result): add new convenience methods

kyranet:


favna (approved):
review not found

favna (dismissed):

packages/result/src/lib/Option/IOption.ts
```diff
@@ -206,6 +223,23 @@ export interface IOption<T> {
...
 
+	/**
+	 * Maps a `None` to the returned `Option<U>` by applying a function to a contained value, leaving `Some<T>`
+	 * untouched.
+	 * @param cb The predicate.
+	 *
+	 * @example
+	 * ```typescript
+	 * const maybeSomeString = some('Hello, world!');
+	 * const maybeSomeLength = maybeSomeString.map((value) => value.length);
```
samfundev, line 234: This example doesn't demonstrate `mapNoneInto`.

packages/result/src/lib/Option/IOption.ts
```diff
@@ -206,6 +223,23 @@ export interface IOption<T> {
...
+	 *
+	 * assert.equal(maybeSomeLength, some(13));
+	 * ```
+	 *
+	 * @note This is an extension not supported in Rust
+	 */
+	mapNoneInto<Inner>(cb: () => Option<Inner>): Option<T | Inner>;
```
samfundev, line 241: I don't understand why this returns a `Option<T | Inner>`. I thought it would just be a `Option<Inner>`.

samfundev (changes requested):

packages/result/src/lib/Option/IOption.ts
kyranet: Because `Some<T>` will map to `Option<T>`, while `None` will map to `Option<Inner>`.

The classes have strict types.

packages/result/src/lib/Option/IOption.ts
kyranet: Resolved in 04f23f4.

favna (approved):

packages/result/src/lib/Option/IOption.ts
samfundev: Wait, would same apply to `mapInto`? I don't think we care about the type of the initial `Some`, it's just about the new `Some`.

PR #406, opened 2022-07-30
feat: add lazy

favna:
Porting this over from Iriss/Nekokai

### DIFF
--- a/packages/utilities/tests/lazy.test.ts
+++ b/packages/utilities/tests/lazy.test.ts
@@ -1,0 +1,15 @@
+import { lazy } from '../src';
+
+describe('lazy', () => {
+	test('GIVEN string callback THEN returns the same', () => {
+		const lazyStoredValue = lazy(() => 'Lorem Ipsum');
+		expect(lazyStoredValue()).toEqual('Lorem Ipsum');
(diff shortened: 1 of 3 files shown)


packages/utilities/tests/lazy.test.ts
```diff
@@ -0,0 +1,15 @@
...
+	test('GIVEN string callback THEN returns the same', () => {
+		const lazyStoredValue = lazy(() => 'Lorem Ipsum');
+		expect(lazyStoredValue()).toEqual('Lorem Ipsum');
+	});
+
+	test('GIVEN string callback with cached value THEN returns the same', () => {
+		const lazyStoredValue = lazy(() => 'Lorem Ipsum');
+		lazyStoredValue();
+		const cachedValue = lazyStoredValue();
+		expect(cachedValue).toEqual('Lorem Ipsum');
```
kyranet, line 13: You need to use `vi.fn` to ensure the callback is called only once.

kyranet (changes requested):
This PR is Not Acceptable, just one change and it'll be good to go.

kyranet (approved):

PR #413, opened 2022-08-04
feat(utilities): add getDeepObjectKeys

favna:
Function I wrote for work and I figured I should port it over, improve the types and write unit tests for it

### DIFF
--- a/packages/utilities/src/lib/getDeepObjectKeys.ts
+++ b/packages/utilities/src/lib/getDeepObjectKeys.ts
@@ -1,0 +1,52 @@
+import { isObject } from './isObject';
+import type { AnyObject } from './utilityTypes';
+
+/**
+ * Flattens an object to a list of its keys, traversing deeply into nested objects and arrays of objects.
+ *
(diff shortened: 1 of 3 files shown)


packages/utilities/src/lib/getDeepObjectKeys.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @param obj The object of which to deeply retrieve its keys
+ * @returns An array of strings holding the keys of the object
+ */
+export function getDeepObjectKeys<T>(
+	obj: AnyObject<T>,
+	{ arrayKeysDottedIndex = true }: GetDeepObjectKeysOptions = { arrayKeysDottedIndex: true }
+): string[] {
+	const keys: string[] = [];
+
+	for (const [key, value] of Object.entries(obj)) {
```
kyranet, line 18: This should support top-level arrays too.

kyranet (changes requested):
To be honest, there should be a private generator function that `yield`s values, it's a lot more efficient than creating an array per level, specially since you can do `yield*` to efficiently defer other generators.

PR #417, opened 2022-08-06
feat(AsyncQueue): add AbortSignal support

kyranet:
~~I hope nobody relies on precise ticks, because this would be a breaking change (`wait()` is 1 tick quicker on non-empty ACs). But to be fair, ECMAScript broke that twice, and [will probably break it a third time](https://github.com/tc39/proposal-faster-promise-adoption). *I hope this wasn't a failed expectation.*~~ 

But basically, this PR implements support for cancellable locks, powered by `AbortSignal`.

As for the weird cast in the `AsyncQueueEntry` file, that's `AbortSignal`'s types being incomplete, at least, in this project.

---

The way the queueing happens now is also different. Before this change, `wait()` waits on the previous item's promise. But because items are now cancellable, doing so would cause the abort to reject the next entry, and because it's chained, this leads to issues. As such, I have come up with a different approach, instead of removing the head when using it, we will keep it, so items that are added while AQ is running are queued and not run immediately.

**tl;dr**:

- **Before**: unlock happens when the previous item resolves. Depends on the previous item's state. Requires promises to always resolve.
- **After**: unlock happens when the current item resolves. Autonomous. Does not put requirements on other items and can dequeue themselves.

---

> **Note**: Keep in mind that the item assigns the abort handlers **only if**:
>
> - The queue was not empty.
> - The AbortSignal wasn't aborted.

> **Note**: `AsyncQueueEntry` is a private API and is not exposed.

---

You can view the tests for examples on how to abort a queue item.

---

### TODO

- [x] Add more documentation (examples on how to abort).
- [x] Add a test to ensure that A -> B -> C, where B is aborted, makes the queue be A -> C.


packages/async-queue/src/lib/AsyncQueue.ts
```diff
@@ -12,21 +14,21 @@ export class AsyncQueue {
...
 	 * async function request(url, options) {
-	 *     await queue.wait();
+	 *     await queue.wait({ signal: options.signal });
 	 *     try {
 	 *         const result = await fetch(url, options);
 	 *         // Do some operations with 'result'
 	 *     } finally {
 	 *         // Remove first entry from the queue and resolve for the next entry
-	 *         queue.shift();
+	 *         queue.unlock();
```
vladfrangu, line 31: Forgotten rename?

vladfrangu (changes requested):

packages/async-queue/src/lib/AsyncQueue.ts
kyranet: Yes - it contained the new code but returned a boolean: whether or not a new item is run. And `shift` would be deprecated calling it.

But then I remembered that AQ is supported in browsers, which drops the possibility to use Node's deprecate utilities, and that I would need to add a test testing a deprecated API or else coverage drops below 100%.

It was just better to leave it as is and keep the PR's scope as small as possible.

favna (approved):

favna:
It's all voodoo magic to me but LGTM I guess

vladfrangu (approved):

PR #430, opened 2022-08-17
feat: add `intoPromise`

kyranet:
This comes very handy for swapping Results and Options with Promises, take the example below:

```typescript
const result = await Result.fromAsync(fetch('https://example.org/image.png'));

return result
	.map(async (response) => {
		const blob = await response.blob();
		return { name: 'image.png', data: await blob.arrayBuffer(), type: blob.type };
	})
	.intoPromise();
```

Which is a lot more idiomatic than:

```typescript
const result = await Result.fromAsync(fetch('https://example.org/image.png'));

return result.match({
	ok: async (response) => {
		const blob = await response.blob();
		return { name: 'image.png', data: await blob.arrayBuffer(), type: blob.type };
	},
	err: (error) => Promise.resolve(err(error))
});
```

### DIFF
--- a/packages/result/src/lib/Option/IOption.ts
+++ b/packages/result/src/lib/Option/IOption.ts
@@ -659,6 +659,19 @@ export interface IOption<T> {
 	flatten<IT>(this: Option<Option<IT>>): Option<IT>;
 
 	/**
+	 * Returns a `Promise` object with the awaited value (if `Some`).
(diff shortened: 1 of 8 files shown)


vladfrangu (approved):

favna (approved):
I don't think I've ran into a situation like described but sure

PR #433, opened 2022-08-23
feat(decorators): silent option to RequiresUserPermissions and RequiresClientPermissions decorator

kaname-png:
This PR is because if we use these decorators for example, in new message events, the log gets full of messages about the bot lacking that permission, etc. The message becomes useless if you use it in listeners since only the developer will see it, so this adds a silence property, which you can configure so that the decorator does not show any error message and only blocks the execution of the assigned method.

BREAKING CHANGES: It is no longer possible to pass a spread as permissions to verify, now an array has to be used.

### DIFF
--- a/packages/decorators/src/djs-decorators.ts
+++ b/packages/decorators/src/djs-decorators.ts
@@ -59,29 +59,30 @@ const DMAvailableUserPermissions = new Permissions(
  * 		return message.channel.send('Showing!');
  * 	}
  *
- * 	(at)RequiresClientPermissions('BAN_MEMBERS') // This subcommand requires the client to be able to ban members.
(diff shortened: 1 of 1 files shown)


favna:
Not too fond of this change at all honestly. Several reasons:

1. This changes the return type of the decorated method in ways that TS is unable to detect, which is unsafe and hard to debug
2. Listeners are inherently not supposed to throw `UserError`s, that class is specific to `Command`s
3. The `run` method should not be decorated, instead you should decorate non-run methods (called from `run`) and handle errors in `run`

If anything we'd have a `@ListenerRequiresClientPermissions`, however as per point 2 above that would be an anti-pattern so I'd rather delegate it to people having it in their own code as opposed to having it in this library which then implicitly encourages the usage.

<details><summary>Click here to see Discord discussion</summary>
image
</details>

kaname-png:
Ok, sounds fair, thanks for the explanation.

PR #458, opened 2022-09-17
feat: add `join` and `joinRaw`

kyranet:
A little utility `lexure` had that we didn't. Most useful for https://github.com/sapphiredev/framework/pull/531.

### DIFF
--- a/packages/lexure/src/lib/util/util.ts
+++ b/packages/lexure/src/lib/util/util.ts
@@ -1,0 +1,39 @@
+import type { Parameter } from '../lexer/streams/ParameterStream';
+
+/**
+ * Joins the parameters by their `leading` value, using the `value` property.
+ * @seealso {@link joinRaw} for the version using `raw` instead of `value`.
+ * @param parameters The parameters to join.
+ * @returns The result of joining the parameters.
+ */
+export function join(parameters: readonly Parameter[]) {
+	if (parameters.length === 0) return '';
+	if (parameters.length === 1) return parameters[0].value;
+
+	let output = parameters[0].value;
+	for (let i = 1; i < parameters.length; i++) {
+		const parameter = parameters[i];
+		output += parameter.leading + parameter.value;
+	}
+
+	return output;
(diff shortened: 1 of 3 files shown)


vladfrangu (approved):

packages/lexure/src/lib/util/util.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export function joinRaw(parameters: readonly Parameter[]) {
+	if (parameters.length === 0) return '';
+	if (parameters.length === 1) return parameters[0].raw;
+
+	let output = parameters[0].raw;
+	for (let i = 1; i < parameters.length; i++) {
+		const parameter = parameters[i];
+		output += parameter.leading + parameter.raw;
+	}
```
favna, line 36:
```suggestion
	for (const parameter of parameters) {
		output += parameter.leading + parameter.raw;
	}
```

Were you half asleep or something

packages/lexure/src/lib/util/util.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export function join(parameters: readonly Parameter[]) {
+	if (parameters.length === 0) return '';
+	if (parameters.length === 1) return parameters[0].value;
+
+	let output = parameters[0].value;
+	for (let i = 1; i < parameters.length; i++) {
+		const parameter = parameters[i];
+		output += parameter.leading + parameter.value;
+	}
```
favna, line 17:
```suggestion
	for (const param of parameters) {
		output += param.leading + param.value;
	}
```

favna (changes requested):

packages/lexure/src/lib/util/util.ts
kyranet: It starts from 1, not 0.

favna: I guess I'm the one half asleep then

favna (approved):

PR #469, opened 2022-09-30
feat(utilities): add `sleep` and `sleepSync`

imranbarbhuiya:
This pr adds `sleep` and `sleepAsync` 

I'm not using `timers/promise` just to make it browser compitible

### DIFF
--- a/packages/utilities/src/index.ts
+++ b/packages/utilities/src/index.ts
@@ -39,7 +39,7 @@ export * from './lib/range';
 export * from './lib/regExpEsc';
 export * from './lib/roundNumber';
 export * from './lib/splitText';
-export { AbortError, sleep, sleepSync, type SleepOptions } from './lib/sleep';
+export * from './lib/sleep';
 export { toTitleCase, ToTitleCaseOptions } from './lib/toTitleCase';
 export * from './lib/tryParse';
 export * from './lib/utilityTypes';
--- a/packages/utilities/src/index.ts
+++ b/packages/utilities/src/index.ts
@@ -39,7 +39,7 @@ export * from './lib/range';
 export * from './lib/regExpEsc';
 export * from './lib/roundNumber';
 export * from './lib/splitText';
-export * from './lib/sleep';
+export { AbortError, sleep, sleepSync, type SleepOptions } from './lib/sleep';
 export { toTitleCase, ToTitleCaseOptions } from './lib/toTitleCase';
 export * from './lib/tryParse';
 export * from './lib/utilityTypes';
--- a/packages/utilities/src/index.ts
+++ b/packages/utilities/src/index.ts
@@ -39,6 +39,7 @@ export * from './lib/range';
 export * from './lib/regExpEsc';
 export * from './lib/roundNumber';
 export * from './lib/splitText';
+export * from './lib/sleep';
 export { toTitleCase, ToTitleCaseOptions } from './lib/toTitleCase';
 export * from './lib/tryParse';
 export * from './lib/utilityTypes';
--- a/packages/utilities/src/lib/sleep.ts
+++ b/packages/utilities/src/lib/sleep.ts
@@ -13,17 +13,16 @@ export interface SleepOptions {
 }
 
 export class AbortError extends Error {
-	public code: string;
+	public readonly code: string;
 	public constructor(
 		message?: string,
 		options?: {
 			cause?: unknown;
-			code?: string;
 		}
 	) {
 		super(message, options);
 		this.name = 'AbortError';
-		this.code = options?.code ?? 'ERR_ABORT';
+		this.code = 'ERR_ABORT';
 	}
 }
 
@@ -36,12 +35,13 @@ export class AbortError extends Error {
 export function sleep<T = undefined>(ms: number, value?: T, options?: SleepOptions): Promise<T> {
 	return new Promise((resolve, reject) => {
 		const timer: NodeJS.Timeout | number = setTimeout(() => resolve(value!), ms);
-		if (options?.signal) {
-			options.signal.addEventListener('abort', () => {
+		const signal = options?.signal;
+		if (signal) {
+			signal.addEventListener('abort', () => {
 				clearTimeout(timer);
 				reject(
 					new AbortError('The operation was aborted', {
-						cause: options.signal!.reason
+						cause: signal.reason
 					})
 				);
 			});
--- a/packages/utilities/src/lib/sleep.ts
+++ b/packages/utilities/src/lib/sleep.ts
@@ -1,21 +1,67 @@
+export interface Abortable {
+	/**
+	 * When provided the corresponding `AbortController` can be used to cancel an asynchronous action.
+	 */
+	signal?: AbortSignal | undefined;
+}
+
+export interface TimerOptions extends Abortable {
+	/**
+	 * Set to `false` to indicate that the scheduled `Timeout`
+	 * should not require the Node.js event loop to remain active.
+	 * @default true
+	 */
+	ref?: boolean | undefined;
+}
+
+export class AbortError extends Error {
+	public code: string;
+	public constructor(
+		message?: string,
+		options?: {
+			cause?: unknown;
+			code?: string;
+		}
+	) {
+		super(message, options);
+		this.name = 'AbortError';
+		this.code = options?.code ?? 'ERR_ABORT';
+	}
+}
+
 /**
  * Sleeps for the specified number of milliseconds.
  * @param ms The number of milliseconds to sleep.
- * @param value The value to resolve the promise with.
+ * @param value A value with which the promise is fulfilled.
  * @see {@link sleepSync} for a synchronous version.
  */
-export function sleep<T = undefined>(ms: number, value?: T): Promise<T> {
-	return new Promise((resolve) => setTimeout(() => resolve(value!), ms));
+export function sleep<T = undefined>(ms: number, value?: T, options?: TimerOptions): Promise<T> {
+	return new Promise((resolve, reject) => {
+		const timer: NodeJS.Timeout | number = setTimeout(() => resolve(value!), ms);
+		if (options?.signal) {
+			options.signal.addEventListener('abort', () => {
+				clearTimeout(timer);
+				reject(
+					new AbortError('The operation was aborted', {
+						cause: new DOMException('The operation was aborted', 'AbortError')
+					})
+				);
+			});
+		}
+		if (options?.ref === false && typeof timer === 'object') {
+			timer.unref();
+		}
+	});
 }
 
 /**
  * Sleeps for the specified number of milliseconds synchronously.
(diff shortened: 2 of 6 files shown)


favna (approved):

packages/utilities/src/lib/sleep.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Sleeps for the specified number of milliseconds synchronously.
+ * @param ms The number of milliseconds to sleep.
+ * @param value The value to return.
+ * @see {@link sleep} for an asynchronous version.
+ */
+export function sleepSync<T = undefined>(ms: number, value?: T): T {
+	const start = Date.now();
+	while (Date.now() - start < ms) continue;
```
kyranet, line 19:
```suggestion
	while (Date.now() - start <= ms) continue;
```

packages/utilities/src/lib/sleep.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Sleeps for the specified number of milliseconds.
+ * @param ms The number of milliseconds to sleep.
+ * @param value The value to resolve the promise with.
+ * @see {@link sleepSync} for a synchronous version.
+ */
+export function sleep<T = undefined>(ms: number, value?: T): Promise<T> {
```
kyranet, line 7: Shouldn't we at least try to support `AbortSignal` like `timers/promises` does?

https://nodejs.org/dist/latest-v18.x/docs/api/timers.html#timerspromisessettimeoutdelay-value-options

Also the option to unref the timeout, of course, although that would be specific to Node.js, since timers in browsers are just an integer rather than a Timer instance. We can just make it do nothing if the timer is a number.

kyranet (changes requested):

packages/utilities/tests/sleep.test.ts
```diff
@@ -0,0 +1,44 @@
...
+			})
+		);
+	});
+});
+
+describe('sleepSync', () => {
+	test('GIVEN a number of ms THEN return after that time', () => {
+		const start = Date.now();
+		sleepSync(1000);
+		expect(Date.now() - start).toBeGreaterThanOrEqual(1000);
```
kyranet, line 33: Please lower this timer, it's... too long for a function that runs hundreds of millions of operations per second. `50` is good.

packages/utilities/tests/sleep.test.ts
```diff
@@ -0,0 +1,44 @@
...
+	});
+
+	test('GIVEN a number of ms and a value THEN return after that time with the value', () => {
+		const start = Date.now();
+		const value = sleepSync(1000, 'test');
+		expect(Date.now() - start).toBeGreaterThanOrEqual(1000);
```
kyranet, line 39: Ditto

packages/utilities/src/lib/sleep.ts
```diff
@@ -0,0 +1,77 @@
...
+		options?: {
+			cause?: unknown;
+			code?: string;
+		}
+	) {
+		super(message, options);
+		this.name = 'AbortError';
+		this.code = options?.code ?? 'ERR_ABORT';
+	}
+}
```
kyranet, line 28: Where does this come from?

packages/utilities/src/lib/sleep.ts
```diff
@@ -0,0 +1,77 @@
...
+		const timer: NodeJS.Timeout | number = setTimeout(() => resolve(value!), ms);
+		if (options?.signal) {
+			options.signal.addEventListener('abort', () => {
+				clearTimeout(timer);
+				reject(
+					new AbortError('The operation was aborted', {
+						cause: new DOMException('The operation was aborted', 'AbortError')
+					})
+				);
+			});
```
kyranet, line 59: We should listen handle the [`AbortSignal.reason`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/reason) property that's set when calling [`AbortController.abort(reason)`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort) and handle it accordingly.

kyranet (changes requested):

packages/utilities/src/lib/sleep.ts
imranbarbhuiya: `timers/promise` trows `AbortError` with cause `DOMException`. So I was trying to make it similar error.

kyranet: Interestingly enough, `timers/promise` sets `cause` as a string, not as a `DOMException`. See below:

https://github.com/nodejs/node/blob/2a4452a53af65a13db4efae474162a7dcfd38dd5/lib/timers/promises.js#L83-L87
https://github.com/nodejs/node/blob/2a4452a53af65a13db4efae474162a7dcfd38dd5/lib/timers/promises.js#L44-L49
https://github.com/nodejs/node/blob/2a4452a53af65a13db4efae474162a7dcfd38dd5/lib/internal/errors.js#L843-L852

And I sorta agree, an exception inside an exception is inefficient and unreadable.

packages/utilities/src/index.ts
```diff
@@ -39,6 +39,7 @@ export * from './lib/range';
 export * from './lib/regExpEsc';
 export * from './lib/roundNumber';
 export * from './lib/splitText';
+export { AbortError, sleep, sleepSync, type SleepOptions } from './lib/sleep';
```
kyranet, line 42: Now that `DOMException` is gone:
```suggestion
export * from './lib/sleep';
```

packages/utilities/src/lib/sleep.ts
```diff
@@ -0,0 +1,65 @@
...
+	public constructor(
+		message?: string,
+		options?: {
+			cause?: unknown;
+			code?: string;
+		}
+	) {
+		super(message, options);
+		this.name = 'AbortError';
+		this.code = options?.code ?? 'ERR_ABORT';
```
kyranet, line 26: You never pass the code:
```suggestion
		this.code = 'ERR_ABORT';
```

packages/utilities/src/lib/sleep.ts
imranbarbhuiya: Yeah after your previous comment, I added `signal.reason` and found that if we don't add a reason, it's automatically `DOMException`.  So changed it to match `timers/promisify`

packages/utilities/src/lib/sleep.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export class AbortError extends Error {
+	public code: string;
+	public constructor(
+		message?: string,
+		options?: {
+			cause?: unknown;
+			code?: string;
+		}
+	) {
```
kyranet, line 23: Following the other review:
```suggestion
	public constructor(message?: string, options?: { cause?: unknown }) {
```

packages/utilities/src/lib/sleep.ts
```diff
@@ -0,0 +1,65 @@
...
+	/**
+	 * Set to `false` to indicate that the scheduled `Timeout`
+	 * should not require the Node.js event loop to remain active.
+	 * @default true
+	 */
+	ref?: boolean | undefined;
+}
+
+export class AbortError extends Error {
+	public code: string;
```
kyranet, line 16: Shouldn't we make this readonly? Or doesn't matter?

packages/utilities/src/lib/sleep.ts
```diff
@@ -0,0 +1,65 @@
...
+ */
+export function sleep<T = undefined>(ms: number, value?: T, options?: SleepOptions): Promise<T> {
+	return new Promise((resolve, reject) => {
+		const timer: NodeJS.Timeout | number = setTimeout(() => resolve(value!), ms);
+		if (options?.signal) {
+			options.signal.addEventListener('abort', () => {
+				clearTimeout(timer);
+				reject(
+					new AbortError('The operation was aborted', {
+						cause: options.signal!.reason
```
kyranet, line 44: You need to assign the signal somewhere, the reason why TS thinks it's nullable is because it's *potentially nullable*.

```ts
const options = { signal: controller.signal };
const promise = sleep(1000, undefined, options);
options.signal = undefined;
controller.abort();
await promise; // <-- ReferenceError!
```

To fix this, assign `options.signal` to a local variable.

kyranet (changes requested):

packages/utilities/src/lib/sleep.ts
```diff
@@ -0,0 +1,65 @@
...
+			});
+		}
+		if (options?.ref === false && typeof timer === 'object') {
+			timer.unref();
+		}
+	});
+}
+
+/**
+ * Sleeps for the specified number of milliseconds synchronously.
```
kyranet, line 56: We should probably note that unlike `sleep` (which uses CPU tick times), `sleepSync` uses wall clock times, so the precision is near-absolute by comparison.

That, and that synchronous means that nothing else in the thread will run for the length of the timer.

packages/utilities/src/lib/sleep.ts
```diff
@@ -0,0 +1,65 @@
...
+ * @param ms The number of milliseconds to sleep.
+ * @param value A value to return.
+ * @see {@link sleep} for an asynchronous version.
+ */
+export function sleepSync<T = undefined>(ms: number, value?: T): T {
+	const start = Date.now();
+	while (Date.now() - start <= ms) continue;
```
kyranet, line 63: Just occurred to me now, why can't we just do this?
```suggestion
	const end = Date.now() + ms;
	while (Date.now() >= end) continue;
```

kyranet (changes requested):

packages/utilities/src/lib/sleep.ts
vladfrangu: Date.now() will only >= end...if its past the wait time, did you mean <= end

imranbarbhuiya: yeah I fixed in while committing

vladfrangu (approved):
LGTM after kyra's suggestion is appplied

packages/utilities/src/lib/sleep.ts
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Sleeps for the specified number of milliseconds synchronously.
+ * @param ms The number of milliseconds to sleep.
+ * @param value A value to return.
+ * @see {@link sleep} for an asynchronous version.
+ */
+export function sleepSync<T = undefined>(ms: number, value?: T): T {
+	const end = Date.now() + ms;
+	while (Date.now() <= end) continue;
```
kyranet, line 63: I somehow swapped the logic earlier, `<=` for `ms: 50` will result on a 51ms wait, we want to break once `Date.now()` equals `end`:
```suggestion
	while (Date.now() < end) continue;
```

kyranet (changes requested):

favna (approved):

vladfrangu (approved):

PR #474, opened 2022-09-30
feat(utilities): add `objectValues` function

### LINKED ISSUE #456: request: add `objectValues` (a strictly typed `Object.values`) to `@sapphire/utilities`

### Feature description

The types built into TypeScript for `Object.values` intentionally aren't fully strictly typed, however there are instances where a strictly typed `Object.values` can be useful. This is where a custom `objectValues` can come in.

### Desired solution

Add a function like this and export it:

```ts
export function objectValues<T extends NonNullObject>(obj: T): T[keyof T][] {
  return Object.values(obj) as T[keyof T][];
}
```

Then implement unit tests for this

### Alternatives considered

N.A.

### Additional context

_No response_

RealShadowNova:
Resolves #456


packages/utilities/tests/objectValues.test.ts
```diff
@@ -0,0 +1,17 @@
+import { objectValues } from '../src';
+
+describe('objectValues', () => {
+	test('GIVEN basic THEN returns expected', () => {
+		const source = { a: 'Hello', b: 420 };
+		const expected = ['Hello', 420];
+
+		expect(objectValues(source)).toEqual(expected);
```
kyranet, line 8: You should at very least type assert the `expect()` call, I believe this would be `expect<string | number>(source)`.

If you do `as const` (please test that readonly works!), it would be `expect<'Hello' | 420>(source)` for enhanced safety.

kyranet (changes requested):

packages/utilities/tests/objectValues.test.ts
RealShadowNova:
```typescript
	test('GIVEN basic THEN returns expected', () => {
		const source = { a: 'Hello', b: 420 } as const;
		const expected = ['Hello', 420] as const;

		expect<('Hello' | 420)[]>(objectValues(source)).toEqual(expected);
	});
```
Like so?

packages/utilities/src/lib/objectValues.ts
```diff
@@ -0,0 +1,5 @@
+import type { NonNullObject } from './utilityTypes';
+
+export function objectValues<T extends NonNullObject>(obj: T): T[keyof T][] {
+	return Object.values(obj) as T[keyof T][];
```
imranbarbhuiya, line 4: The typing isn't correct for array
image

imranbarbhuiya (changes requested):

packages/utilities/src/lib/objectValues.ts
imranbarbhuiya, line 4:
```suggestion
export function objectValues<T extends NonNullObject>(obj: T): T extends ArrayLike<any> ? T[number][] : T[keyof T][] {
	return Object.values(obj) as T extends ArrayLike<any> ? T[number][] : T[keyof T][]
```
works

imranbarbhuiya (changes requested):

packages/utilities/tests/objectValues.test.ts
kyranet: Yes, although `expected` doesn't need `as const`, and we should have a type without readonly types.

imranbarbhuiya (dismissed):

favna (approved):

packages/utilities/src/lib/objectValues.ts
```diff
@@ -0,0 +1,5 @@
+import type { NonNullObject } from './utilityTypes';
+
+export function objectValues<T extends NonNullObject>(obj: T): T extends ArrayLike<any> ? T[number][] : T[keyof T][] {
```
vladfrangu, line 3: Shouldn't this use `T extends ArrayLike<infer Values> ? Values[] : T[keyof T][]`?

vladfrangu: I guess both work

vladfrangu (approved):

packages/utilities/src/lib/objectValues.ts
```diff
@@ -0,0 +1,5 @@
...
+	return Object.values(obj) as T extends ArrayLike<any> ? T[number][] : T[keyof T][];
```
favna, line 4:
```suggestion
	return Object.values(obj) as T extends ArrayLike<infer Values> ? Values[] : T[keyof T][];
```

favna (changes requested):

packages/utilities/src/lib/objectValues.ts
RealShadowNova: What is `Values` in this context?

PR #475, opened 2022-09-30
feat(result): add `Result.unwrapRaw`

favna:
This is essentially a shortcut for:
```ts
	return result.match({
		err: (error) => {
			throw error;
		},
		ok: (value) => value.data[0]
	});
```
which can then be written as:
```ts
return result.map((value) => value.data[0]).unwrapRaw();
```

This is useful when instead of throwing a `ResultError` you want to instead throw the inner error (i.e. `ArgumentError` for Sapphire, see sapphiredev/framework#528, or a `FetchError` from @skyra/safe-fetch, something that brought about this change)

### DIFF
--- a/packages/result/src/lib/Result/IResult.ts
+++ b/packages/result/src/lib/Result/IResult.ts
@@ -532,6 +532,31 @@ export interface IResult<T, E> {
 	unwrapOrElse<V>(op: (error: E) => V): T | V;
 
 	/**
+	 * Returns the contained `Ok` value.
+	 *
+	 * If the value is an `Err`, it throws the contained error.
+	 * @seealso {@link unwrap}
+	 * @seealso {@link unwrapOr}
+	 * @seealso {@link unwrapOrElse}
+	 *
+	 * @example
+	 * ```typescript
+	 * const x = ok(2);
+	 * assert.equal(x.unwrapRaw(), 2);
+	 * ```
+	 * @example
+	 * ```typescript
+	 * const x = err('Emergency failure');
+	 * assert.throws(() => x.unwrapRaw(), {
+	 *   name: 'Error',
+	 *   message: 'Unwrap failed',
+	 *   value: 'Emergency failure'
+	 * });
+	 * ```
+	 */
+	unwrapRaw(): T | never;
+
+	/**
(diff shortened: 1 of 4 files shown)


packages/result/tests/Result.test.ts
```diff
@@ -438,6 +438,28 @@ describe('Result', () => {
...
 
+		describe('unwrapRaw', () => {
+			test('GIVEN ok THEN returns value', () => {
+				const x = ok(2);
+
+				expect<number>(x.unwrapRaw()).toBe(2);
+			});
+
+			test('GIVEN err THEN throws Error', () => {
+				const x = err(new Error('Some error message'));
```
kyranet, line 449: Assign the error to a variable.

packages/result/tests/Result.test.ts
```diff
@@ -438,6 +438,28 @@ describe('Result', () => {
...
+
+				try {
+					x.unwrapRaw();
+				} catch (raw) {
+					const error = raw as Error;
+
+					expect(error).toBeInstanceOf(Error);
+					expect<string>(error.name).toBe('Error');
+					expect<string>(error.message).toBe('Some error message');
+				}
```
kyranet, line 459:
```suggestion
				const error = new Error('Some error message');
				const x = err(error);

				expect(() => x.unwrapRaw()).toThrowError(error);
```

packages/result/src/lib/Result/IResult.ts
```diff
@@ -531,6 +531,31 @@ export interface IResult<T, E> {
 	 */
 	unwrapOrElse<V>(op: (error: E) => V): T | V;
 
+	/**
+	 * Returns the contained `Ok` value.
+	 *
+	 * If the value is an `Err`, it throws the contained error.
+	 * @seealso {@link unwrap}
+	 * @seealso {@link unwrapOr}
+	 * @seealso {@link unwrapOrElse}
```
kyranet, line 540: The linked unwrap methods should `@seealso {@link unwrapRaw}`.

kyranet (changes requested):

kyranet (approved):

vladfrangu (approved):

Issue #481, opened 2022-10-03
request: add `tryParseURL` to utilities + rename `tryParse` to `tryParseJSON` and add deprecation notice to the former of those two.


favna:
### Is there an existing issue or pull request for this?

- [X] I have searched the existing issues and pull requests

### Feature description

We have a `tryParse` in /utilities which wraps `JSON.parse` in a try/catch. Another case where you may want to parse and catch is `new URL(...)` so lets introduce `tryParseURL`

### Desired solution

- Implement new file
- Do something smart for handling the type of `URL`. We cannot rely on people having DOM in their lib. Probably a triple slash directive.
- Add new unit tests
- Rename `tryParse` to `tryParseJSON`
- Add an alias for `tryParseJSON` with `@deprecated` a TSDoc comment
- Add the alias to the aliases map for subpath exports
- Run `check-subpath` in `packages/utilities` to update the package.json

```ts
/**
 * Tries parse a string to a {@link URL} object
 * @param value The possible URL to parse
 * @returns an URL object if it was a valid URL or `null` if it was not.
 */
export function tryParseURL(value: string): URL | null {
  try {
    return new URL(value);
  } catch {
    return null;
  }
}
```

### Alternatives considered

Not implementing it? Jk

legendhimslef:
I am working on it.

PR #482, opened 2022-10-03
feat: added `tryParseURL`

legendhimslef:
fixes: #481 
Todo:
- [x] Write unit test


packages/utilities/package.json
```diff
@@ -262,9 +262,19 @@
 			"require": "./dist/lib/toTitleCase.js"
 		},
 		"./tryParse": {
```
vladfrangu, line 264: Should probably remove this

packages/utilities/src/lib/tryParseJSON.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Try parse a stringified JSON string.
+ * @param value The value to parse
+ */
+export function tryParseJSON(value: string): object | string {
```
vladfrangu, line 5: Should return whatever json.parse says it returns in typings

packages/utilities/src/lib/tryParseJSON.ts
```diff
@@ -0,0 +1,20 @@
...
+	try {
+		return JSON.parse(value);
+	} catch (err) {
+		return value;
+	}
+}
+
+/**
```
vladfrangu, line 13: @favna do we need to do this or can we just axe it ?

vladfrangu (changes requested):

packages/utilities/package.json
imranbarbhuiya: it'll be a breaking change ig?

packages/utilities/src/lib/tryParseJSON.ts
imranbarbhuiya: `JSON.parse` return type is `any`

packages/utilities/src/lib/tryParseJSON.ts
favna: Yes, removing functions is a breaking change

packages/utilities/package.json
favna: Indeed, removing functions is a breaking change

favna (approved):
I'll be honest I had completely missed that we already had `parseURL` because I was looking for `try...` when I created the issue LMAO!! 

Ohwell, ultimately we also got general improvements out of this PR

imranbarbhuiya:
@favna I was just checking if there is any way to add deprecated jsdoc in re-export and found it 

https://github.com/microsoft/TypeScript/issues/42684

favna:
> @favna I was just checking if there is any way to add deprecated jsdoc in re-export and found it
> 
> [microsoft/TypeScript#42684](https://github.com/microsoft/TypeScript/issues/42684)

Oh can you make a PR for that then? Same for `filterNullishOrZero` and `filterNullishOrEmpty` (mind the OR, so not the AND variants)

PR #483, opened 2022-10-03
feat: throttle function

legendhimslef:


packages/utilities/tests/throttle.test.ts
```diff
@@ -0,0 +1,26 @@
...
+	test('GIVEN number callback THEN returns the same output until the delay', async () => {
+		const callback = vi.fn((num: number) => num);
+
+		const throttleFunc = throttle(callback, 100);
+		const now = Date.now();
+		expect(throttleFunc(now)).toEqual(now);
+		expect(throttleFunc(100)).toEqual(now);
+		expect(callback).toHaveBeenCalledOnce();
+		await sleep(100);
+		expect(throttleFunc(200)).toEqual(200);
```
kyranet, line 13: This is flaky, please let the throttle wait be at least half the time of the sleep to leave +50ms variance.

packages/utilities/src/lib/throttle.ts
```diff
@@ -0,0 +1,31 @@
...
+			}
+
+			return prevValue;
+		},
+		{
+			flush() {
+				prev = 0;
+			}
+		}
+	) as T & { flush: () => void };
```
kyranet, line 30: Can't we make a type named `ThrottleFn`? It'll help reducing the type duplication.

kyranet (changes requested):
I don't know how to feel about that `Object.assign`, but I guess it works.

favna (approved):

kyranet (approved):

PR #484, opened 2022-10-05
fix(utilities): race condition issue for closed paginated message

goestav:
This pull request fixes a bug where the button components won't get removed due to a race condition error: `DiscordAPIError: Interaction has already been acknowledged.`.
This occurs when the user tries to interact with the message when the paginated message is about to close.
You can reproduce this by spam clicking the next/previous buttons on a paginated message.


favna:
dunno what you see as spam clicking because it either works for me, or I get rate limited by Discord 

https://user-images.githubusercontent.com/4019718/194717498-bece53a8-ec95-437d-9743-af6f89a6459b.mp4

favna:
In general the code looks good but please attach a reproduction video recording because I cannot reproduce it

goestav:
https://user-images.githubusercontent.com/27970303/194729085-cf2c0937-5060-4c40-87e5-fa222a63e5c7.mov

I guess I didn't explain well, but it happens while the paginated message is getting closed
In this video you can see that inside the `handleEnd` method it's still on the previous page (4th page in debugger while it's the 5th page on discord) and it throws a DiscordAPIError because of this
The components also won't get removed

favna (approved):

PR #486, opened 2022-10-08
feat(utilities): add pickRandom

Lioness100:


packages/utilities/src/lib/pickRandom.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Picks a random element from an array
+ * @param array The array to pick a random element from
+ */
+export function pickRandom<T>(array: readonly T[]): T {
```
imranbarbhuiya, line 5: can you add a `length` option that'll default to 1? So, 

```ts
pickRandom([1,2,3,4], 1); 
// 2

pickRandom([1,2,3,4], 2);
// 3,2
```

imranbarbhuiya (changes requested):

packages/utilities/tests/pickRandom.test.ts
```diff
@@ -0,0 +1,21 @@
...
+
+	test('GIVEN count of one THEN picks one random element', () => {
+		const array = ['a', 'b', 'c'];
+		expect(array).toContain(pickRandom(array, 1));
+	});
+
+	test('GIVEN count of two THEN picks two random elements', () => {
+		const array = ['a', 'b', 'c'];
+		const picked = pickRandom(array, 2);
+		expect(array).toContain(picked[0]);
```
imranbarbhuiya, line 17: You can use https://vitest.dev/api/#expect-arraycontaining too

imranbarbhuiya (changes requested):

packages/utilities/tests/pickRandom.test.ts
Lioness100: I think that has to be used in an equality check

imranbarbhuiya:
```ts
expects(array).toEquals(expects.arrayContaining(pickRandom(array, 2)));
```

Lioness100: Oh, I see. Yeah sure

packages/utilities/src/lib/pickRandom.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Picks a random element from an array
+ * @param array The array to pick a random element from
+ */
+export function pickRandom<T>(array: readonly T[], count?: 1): T;
+export function pickRandom<T>(array: readonly T[], count: number): T[];
+export function pickRandom<T>(array: readonly T[], count = 1): T | T[] {
```
imranbarbhuiya, line 7: Sorry for not saying this in 1 review, but I realized this method can return the same elements multiple times. Maybe doing something like this will be a better approach
https://github.com/discordjs/discord.js/blob/main/packages/collection/src/collection.ts#L179

imranbarbhuiya (changes requested):

packages/utilities/src/lib/pickRandom.ts
favna: Agreed. And sticking to their impl is a good idea anyway as that's been pretttyyyyy much proven to be good.

favna (approved):

packages/utilities/src/lib/pickRandom.ts
```diff
@@ -0,0 +1,12 @@
...
+	const arr = [...array];
+	if (typeof count === 'undefined') return arr[Math.floor(Math.random() * arr.length)];
```
imranbarbhuiya, line 9:
```suggestion
	if (typeof count === 'undefined' || count === 1) return arr[Math.floor(Math.random() * arr.length)];
```
or change the typings

imranbarbhuiya (changes requested):

packages/utilities/src/lib/pickRandom.ts
Lioness100: I'm pushing a typing change

packages/utilities/src/lib/pickRandom.ts
```diff
@@ -1,12 +1,20 @@
 /**
  * Picks a random element from an array
  * @param array The array to pick a random element from
+ * @param amount - Amount of values to obtain randomly
```
Lioness100, line 4: Inconsistent dash after "amount"

favna (approved):

PR #492, opened 2022-10-08
refactor(duration): reuse Time enum

Lioness100:
The Time enum is already exported in `constants.ts`. It's worth noting that `/timestamp` also exports the same enum, but I think it would be a bit overkill to create another package just to share one small enum. It does bring up the question of whether the packages usages should copy and paste the same `Time` documentation?


favna:
yes copy the docs, no not a package for 1 enum.

favna (approved):

PR #498, opened 2022-10-09
refactor(utilities): improve `codeBlock` and `inlineCodeBlock`

favna:
Now, these functions are more similar to their sister functions in DiscordJS in terms of typing while retaining the additional safety checks that Sapphire has always provided.


vladfrangu (dismissed):

packages/utilities/src/lib/codeBlock.ts
```diff
@@ -1,14 +1,27 @@
...
+	if (typeof content === 'string') {
+		if (content.length === 0) {
+			return `\`\`\`\n${zws}\n\`\`\``;
+		}
+
+		return `\`\`\`${language}\n${content.replace(/```/, `\`${zws}\`\``).replace(/`$/g, `\`${zws}`)}\n\`\`\``;
 	}
-	return `\`\`\`${language}\n${expression || zws}\`\`\``;
+
+	return typeof content === 'undefined' ? `\`\`\`\n${language}\n\`\`\`` : `\`\`\`${language}\n${content || zws}\n\`\`\``;
```
kyranet, line 26: Is this branch even possible? We already have line 18 for this.

kyranet (changes requested):

packages/utilities/src/lib/codeBlock.ts
favna: If someone in JS writes something like:
```ts
const myCode = someFunc(); // Lets say this returns undefined for whatever reason

codeBlock('js', myCode);
```
then line 18 will be false

So not necessarily

kyranet: That's literally what line 18 does, it already covers the `string` case.

favna: `myCode` in my example is undefined so it's NOT covered by the `string` case

vladfrangu (approved):

PR #502, opened 2022-10-14
feat: add `@sapphire/bitfield` package

kyranet:
## TODO

- [x] Write tests
- [x] Write README
- [x] Add references to the new package in the rest of the repo

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Install Dependencies
         run: yarn --immutable
       - name: Build test dependencies
-        run: yarn turbo run build --filter=@sapphire/async-queue --filter=@sapphire/bitfield --filter=@sapphire/*utilities --filter=@sapphire/duration --filter=@sapphire/fetch --filter=@sapphire/lexure --filter=@sapphire/ratelimits --filter=@sapphire/result --filter=@sapphire/timer-manager  --filter=@sapphire/stopwatch --no-deps
+        run: yarn turbo run build --filter="@sapphire/async-queue" --filter="@sapphire/*utilities" --filter="@sapphire/duration" --filter="@sapphire/fetch" --filter="@sapphire/lexure" --filter="@sapphire/ratelimits" --filter="@sapphire/result" --filter="@sapphire/timer-manager"  --filter="@sapphire/stopwatch" --no-deps
       - name: Run tests
         run: yarn test
       - name: Store code coverage report
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Install Dependencies
         run: yarn --immutable
       - name: Build test dependencies
-        run: yarn turbo run build --filter=@sapphire/async-queue --filter=@sapphire/*utilities --filter=@sapphire/duration --filter=@sapphire/fetch --filter=@sapphire/lexure --filter=@sapphire/ratelimits --filter=@sapphire/result --filter=@sapphire/timer-manager  --filter=@sapphire/stopwatch --no-deps
+        run: yarn turbo run build --filter=@sapphire/async-queue --filter=@sapphire/bitfield --filter=@sapphire/*utilities --filter=@sapphire/duration --filter=@sapphire/fetch --filter=@sapphire/lexure --filter=@sapphire/ratelimits --filter=@sapphire/result --filter=@sapphire/timer-manager  --filter=@sapphire/stopwatch --no-deps
       - name: Run tests
         run: yarn test
       - name: Store code coverage report
--- a/packages/bitfield/README.md
+++ b/packages/bitfield/README.md
@@ -216,10 +216,10 @@ PermissionsBitField.toObject(PermissionFlags.Read | PermissionFlags.Write);
 PermissionsBitField.toObject([PermissionFlags.Read, PermissionFlags.Write]);
 PermissionsBitField.toObject(['Read', 'Write']);
 // {
-// 	Read: true,
-// 	Write: true,
-// 	Edit: false,
-// 	Delete: false
+//   Read: true,
+//   Write: true,
+//   Edit: false,
+//   Delete: false
 // }
 ```
 
--- a/packages/bitfield/README.md
+++ b/packages/bitfield/README.md
@@ -27,7 +27,7 @@
     -   [Getting the symmetric difference between different fields](#getting-the-symmetric-difference-between-different-fields)
     -   [Formatting a field](#formatting-a-field)
 -   [Buy us some doughnuts](#buy-us-some-doughnuts)
--   [Contributors ✨](#contributors-%E2%9C%A8)
+-   [Contributors ✨](#contributors-)
 
 ## Features
 
@@ -49,7 +49,7 @@ npm install @sapphire/bitfield
 **Note:** While this section uses `require`, the imports match 1:1 with ESM imports. For example `const { BitField } = require('@sapphire/bitfield')` equals `import { BitField } from '@sapphire/bitfield'`.
 
 ```typescript
-// Import the BitField class
+// Require the BitField class
 const { BitField } = require('@sapphire/bitfield');
 
 const PermissionFlags = {
--- a/packages/bitfield/README.md
+++ b/packages/bitfield/README.md
@@ -1,0 +1,125 @@
+<div align="center">
+
+![Sapphire Logo](https://cdn.skyra.pw/gh-assets/sapphire-banner.png)
+
+# @sapphire/bitfield
+
+**BitField utilities for JavaScript.**
+
+[![GitHub](https://img.shields.io/github/license/sapphiredev/utilities)](https://github.com/sapphiredev/utilities/blob/main/LICENSE.md)
+[![codecov](https://codecov.io/gh/sapphiredev/utilities/branch/main/graph/badge.svg?token=OEGIV6RFDO)](https://codecov.io/gh/sapphiredev/utilities)
+[![npm bundle size](https://img.shields.io/bundlephobia/min/@sapphire/bitfield?logo=webpack&style=flat-square)](https://bundlephobia.com/result?p=@sapphire/bitfield)
+[![npm](https://img.shields.io/npm/v/@sapphire/bitfield?color=crimson&logo=npm&style=flat-square)](https://www.npmjs.com/package/@sapphire/bitfield)
+
+</div>
+
+**Table of Contents**
+
+-   [Features](#features)
+-   [Installation](#installation)
+-   [Usage](#usage)
+-   [Buy us some doughnuts](#buy-us-some-doughnuts)
+-   [Contributors ✨](#contributors-%E2%9C%A8)
+
+## Features
+
+-   Written in TypeScript
+-   Bundled with esbuild so it can be used in NodeJS and browsers
+-   Offers CommonJS, ESM and UMD bundles
+-   Fully tested
+
+## Installation
+
+You can use the following command to install this package, or replace `npm install` with your package manager of choice.
+
+```sh
+npm install @sapphire/bitfield
+```
+
+## Usage
+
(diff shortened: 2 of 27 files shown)


packages/bitfield/src/lib/BitField.ts
```diff
@@ -0,0 +1,279 @@
...
+		}
+
+		const entries = Object.entries(flags) as [string, Flags[keyof Flags]][];
+		if (entries.length === 0) {
+			throw new TypeError('flags must be a non-empty object');
+		}
+
+		const type = typeof entries[0][1];
+		if (type !== 'number' && type !== 'bigint') {
+			throw new TypeError('type must be either "number" or "bigint"');
```
vladfrangu, line 19: Could have a better error message

packages/bitfield/src/lib/BitField.ts
```diff
@@ -0,0 +1,279 @@
...
+	 * @example
+	 * ```typescript
+	 * const bitfield = new BitField({
+	 * 	Read:   0b0001,
+	 * 	Write:  0b0010,
+	 * 	Edit:   0b0100,
+	 * 	Delete: 0b1000
+	 * });
+	 *
+	 * bitfield.complement(0b0101);
```
vladfrangu, line 251:
```suggestion
	 * bitfield.toObject(0b0101);
```

packages/bitfield/tests/lib/BitField.test.ts
```diff
@@ -0,0 +1,355 @@
...
+
+describe('BitField', () => {
+	const NumberFlags = {
+		Read: 1,
+		Write: 2,
+		Edit: 4,
+		Delete: 8
+	} as const;
+
+	const BigIntFlags = {
```
vladfrangu, line 11: Can you make sure enums work too? 🙏

vladfrangu (changes requested):

packages/bitfield/src/lib/BitField.ts
kyranet: What do you suggest?

packages/bitfield/tests/lib/BitField.test.ts
kyranet: I don't know if enums will work, or if I can map their types nicely, but it can be implemented in a future PR.

2-way enums, that is. 1-way ones should be fine.

packages/bitfield/src/lib/BitField.ts
vladfrangu: `A bitfield can only use numbers or bigints for its values` or something... Also you really should do that check for every entry provided by the user

packages/bitfield/tests/lib/BitField.test.ts
vladfrangu: 1. Check out how we did it in shapeshift's nativeenum

packages/bitfield/src/lib/BitField.ts
kyranet: It's done later, lines 25-48, does specific checks to ensure all values are the same type as the first one.

packages/bitfield/tests/lib/BitField.test.ts
kyranet: The biggest issue for me is mapping the values correctly while ignoring the reverse maps, both in the types and in the data.

Chances are high I'll just make a function to transform 2-way enums into 1-way ones, otherwise, if done within the constructor, the types could get potentially A LOT more complex, and it's already nightmare-ish as-is.

vladfrangu (approved):

.github/workflows/continuous-integration.yml
```diff
@@ -47,7 +47,7 @@ jobs:
       - name: Install Dependencies
         run: yarn --immutable
       - name: Build test dependencies
-        run: yarn turbo run build --filter=@sapphire/async-queue --filter=@sapphire/*utilities --filter=@sapphire/duration --filter=@sapphire/fetch --filter=@sapphire/lexure --filter=@sapphire/ratelimits --filter=@sapphire/result --filter=@sapphire/timer-manager  --filter=@sapphire/stopwatch --no-deps
+        run: yarn turbo run build --filter=@sapphire/async-queue --filter=@sapphire/bitfield --filter=@sapphire/*utilities --filter=@sapphire/duration --filter=@sapphire/fetch --filter=@sapphire/lexure --filter=@sapphire/ratelimits --filter=@sapphire/result --filter=@sapphire/timer-manager  --filter=@sapphire/stopwatch --no-deps
```
favna, line 50:
```suggestion
        run: yarn turbo run build --filter="@sapphire/async-queue" --filter="@sapphire/*utilities" --filter="@sapphire/duration" --filter="@sapphire/fetch" --filter="@sapphire/lexure" --filter="@sapphire/ratelimits" --filter="@sapphire/result" --filter="@sapphire/timer-manager"  --filter="@sapphire/stopwatch" --no-deps
```

bitfield only has to be specified here if other packages, or their dependencies, directly depend on bitfield which would then cause type checking errors when tests are ran. Seeing as that is not the case, we don't need to specify it here. 

Also wrapped the packages in quotes to finally get rid of the annoyance where this command cannot be simply copy pasted to a terminal to test because of `*utilities` <_<

packages/bitfield/README.md
```diff
@@ -0,0 +1,306 @@
...
+```sh
+npm install @sapphire/bitfield
+```
+
+## Usage
+
+**Note:** While this section uses `require`, the imports match 1:1 with ESM imports. For example `const { BitField } = require('@sapphire/bitfield')` equals `import { BitField } from '@sapphire/bitfield'`.
+
+```typescript
+// Import the BitField class
```
favna, line 52:
```suggestion
// Require the BitField class
```

packages/bitfield/src/lib/BitField.ts
favna: Line 19 is not wrapped in a try-catch so this particular error can still be thrown. Please adjust the error message as requested by @vladfrangu

packages/bitfield/src/lib/BitField.ts
```diff
@@ -0,0 +1,279 @@
...
+	 * // 0b1011
+	 * ```
+	 */
+	public complement(field: ValueResolvable<this>): ValueType<this> {
+		return this.difference(this.mask, field);
+	}
+
+	/**
+	 * Makes the union of all bits.
+	 * @param fields The bits to create an union of.
```
favna, line 120:
```suggestion
	 * Makes a union of all the bits.
	 * @param fields The bits to create a union of.
```

packages/bitfield/src/lib/BitField.ts
```diff
@@ -0,0 +1,279 @@
...
+		}
+
+		return field;
+	}
+
+	/**
+	 * Makes the intersection of all bits.
+	 * @param bitfield The first field.
+	 * @param fields The bits to intersect with `bitfield`.
+	 * @returns The result of intersecting `bitfield` with all of the `fields`, also denoted as `A ⋂ fields`.
```
favna, line 145:
```suggestion
	 * Makes an intersection of all the bits.
	 * @param bitfield The first field.
	 * @param fields The bits to intersect with `bitfield`.
	 * @returns The result of intersecting `bitfield` with all of the `fields`, also denoted as `A ⋂ fields`.
```

packages/bitfield/src/lib/BitField.ts
```diff
@@ -0,0 +1,279 @@
...
+		let field = this.resolve(bitfield);
+		for (const resolvable of fields) {
+			field = (field & this.resolve(resolvable)) as ValueType<this>;
+		}
+
+		return field;
+	}
+
+	/**
+	 * Removes from `a` the bits from `b`.
```
favna, line 169:
```suggestion
	 * Removes from `a` the bits that exist in `b`.
```

packages/bitfield/src/lib/BitField.ts
```diff
@@ -0,0 +1,279 @@
...
+	 * @param a The first field.
+	 * @param b The bits to create an union of.
```
favna, line 171: This appears to be copypasted from union

packages/bitfield/tsup.config.ts
```diff
@@ -0,0 +1,3 @@
+import { createTsupConfig } from '../../scripts/tsup.config';
+
+export default createTsupConfig({ globalName: 'SapphireBitField', target: 'es2020' });
```
favna, line 3:
```suggestion
export default createTsupConfig({ globalName: 'SapphireBitField' });
```

We can target ES2021

packages/bitfield/tests/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+	"extends": "../../ts-config/extra-strict-without-decorators.json",
+	"compilerOptions": {
+		"target": "ES2020",
```
favna, line 4:
```suggestion
```

packages/bitfield/vitest.config.ts
```diff
@@ -0,0 +1,7 @@
+import { createVitestConfig } from '../../scripts/vitest.config';
+
+export default createVitestConfig({
+	esbuild: {
+		target: 'es2020'
+	}
+});
```
favna, line 7:
```suggestion
export default createVitestConfig();
```

turbo.json
```diff
@@ -16,6 +16,7 @@
 		"test": {
 			"dependsOn": [
 				"@sapphire/async-queue#build",
+				"@sapphire/bitfield#build",
```
favna, line 19:
```suggestion
```

favna (changes requested):

packages/bitfield/src/lib/BitField.ts
```diff
@@ -0,0 +1,279 @@
...
+	public has(field: ValueResolvable<this>, bits: ValueResolvable<this>): boolean {
+		const resolved = this.resolve(bits);
+		return (this.resolve(field) & resolved) === resolved;
+	}
+
+	/**
+	 * Makes the complement of `field`, which is a field of all bits (of `U` or the union of all {@link Flags} bits)
+	 * that do not belong to `A`. It is the result of `U ∖ A`, or `difference(U, field)`.
+	 * @param field The bits to get the complement of.
+	 * @returns The complement of `field`, also denoted `Aᶜ` or `A'`.
```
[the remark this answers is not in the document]
kyranet: There's only one complement, that's why I called it "the" complement instead of "a" complement.

packages/bitfield/src/lib/BitField.ts
kyranet: It's a leftover from an older version of the code, `b` used to be a rest parameter that called `this.union(...b)`. Needs changing, yeah.

packages/bitfield/src/lib/BitField.ts
kyranet: I... already did.

packages/bitfield/src/lib/BitField.ts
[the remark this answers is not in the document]
favna: This isn't about singular or plural but about English sentence structure. You always make "a complement". You'll see in articles such as https://www.grammarly.com/blog/complement-compliment/ (which admittedly do not address "the" vs "a" specifically but still apply) and https://dictionary.cambridge.org/dictionary/english/complement that they always write "a complement"

Furthermore the plural of "complement" is "complements" so if this was a matter of pluralization then "the complement" would be wrong as well as it would have to have been "the complements". That is not the case here however.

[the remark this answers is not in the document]
kyranet: We're talking about mathematical terms here, Wikipedia refers it as "the" complement, 13 times.

https://en.wikipedia.org/wiki/Complement_(set_theory)

packages/bitfield/src/lib/BitField.ts
favna:
```suggestion
			throw new TypeError('A bitfield can only use `number`s or `bigint`s for its values.');
```

This is all Vladdy asked. Was that so hard...

favna (approved):

PR #531, opened 2022-12-24
feat(snowflake): added `Snowflake.compare`

kyranet:
Useful for sorting structures by snowflakes, also does string length comparison when both values are strings to avoid false positives, as `a < b` (as both strings) will do an alphabetical order sort rather than a number sort.

### DIFF
--- a/packages/snowflake/src/lib/Snowflake.ts
+++ b/packages/snowflake/src/lib/Snowflake.ts
@@ -106,6 +106,29 @@ export class Snowflake {
 	public timestampFrom(id: string | bigint): number {
 		return Number((BigInt(id) >> 22n) + this.#epoch);
 	}
+
+	/**
+	 * Returns a number indicating whether a reference snowflake comes before, or after, or is same as the given
+	 * snowflake in sort order.
+	 * @param a The first snowflake to compare.
+	 * @param b The second snowflake to compare.
+	 * @returns `-1` if `a` is older than `b`, `0` if `a` and `b` are equals, `1` if `a` is newer than `b`.
+	 * @example
+	 * ```typescript
+	 * const ids = ['737141877803057244', '1056191128120082432', '254360814063058944'];
+	 * console.log(ids.sort(a, b) => Snowflake.compare(a, b));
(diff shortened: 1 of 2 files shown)


packages/snowflake/src/lib/Snowflake.ts
```diff
@@ -106,6 +106,29 @@ export class Snowflake {
...
+	/**
+	 * Returns a number indicating whether a reference snowflake comes before, or after, or is same as the given
+	 * snowflake in sort order.
+	 * @param a The first snowflake to compare.
+	 * @param b The second snowflake to compare.
+	 * @returns `-1` if `a` is older than `b`, `0` if `a` and `b` are equals, `1` if `a` is newer than `b`.
+	 * @example
+	 * ```typescript
+	 * const ids = ['737141877803057244', '1056191128120082432', '254360814063058944'];
+	 * console.log(ids.sort(a, b) => Snowflake.compare(a, b));
```
favna, line 119:
```suggestion
	 * console.log(ids.sort(Snowflake.compare));
```

favna (changes requested):

packages/snowflake/src/lib/Snowflake.ts
kyranet: ESLint whines about `this` being unbound, which is why I didn't write it like that. The solution, `this: void`, arises another ESLint error (invalid void) and doesn't silence the existing one.

Also, having the call done manually helps figuring out how to do DESC sort rather than ASC, which is basically negated calls.

favna: Doesn't take away that the old code is invalid JavaScript???

kyranet: I literally ran them right now, both examples work and produce the correct output as I wrote them in each `@example` block:

image

favna (approved):

Syjalo:
Type of `a` and `b` can't be string and BigInt in the same time. The if statement is useless. Also if both `a` and `b` will be a string, only `a` will be converted to a BigInt.

PR #561, opened 2023-03-10
feat(duration): add units on Duration

kyranet:


packages/duration/src/lib/constants.ts
```diff
@@ -1,13 +1,16 @@
...
-	Day = 1000 * 60 * 60 * 24,
-	Month = 1000 * 60 * 60 * 24 * (365 / 12),
-	Year = 1000 * 60 * 60 * 24 * 365
+	Minute = Second * 60,
+	Hour = Minute * 60,
+	Day = Hour * 24,
+	Week = Day * 7,
+	Month = Day * (365 / 12),
+	Year = Day * 365
 }
```
favna, line 14: This enum also exists in /cron and /timestamp, update it there as well. Do so in a separate PR and separate commits so they it can rebased & merged and each commit only touches 1 package.

packages/duration/src/lib/Duration.ts
```diff
@@ -111,12 +143,44 @@ export class Duration {
...
+
+	/**
+	 * The RegExp used for removing commas
+	 */
+	private static readonly kCommaRegex = /,/g;
+
+	/**
+	 * The RegExp used for replacing a/an with 1
+	 */
+	private static readonly kAanRegex = /\ban?\b/gi;
```
favna, line 185: Didn't know it was 2019 with these `k` prefixes for `klasa`

favna (changes requested):

packages/duration/tests/lib/Duration.test.ts
```diff
@@ -1,14 +1,85 @@
...
+			expect(duration.days).toEqual(units.days ?? 0);
+			expect(duration.weeks).toEqual(units.weeks ?? 0);
+			expect(duration.months).toEqual(units.months ?? 0);
+			expect(duration.years).toEqual(units.years ?? 0);
+		}
 
-	test('GIVEN duration with an offset of 1s, THEN shows 1000ms', () => {
-		const duration = new Duration('a second');
-		expect(duration.offset).toEqual(1000);
+		test.each(['0ns', '0μs', '0ms', '0s', '0m', '0h', '0d', '0w', '0mo', '0yr'])('GIVEN %s THEN shows 0ms', (pattern) => {
```
vladfrangu, line 18: lets match `0us` as `0microseconds` too

packages/duration/src/lib/constants.ts
```diff
@@ -1,13 +1,16 @@
...
-	Minute = 1000 * 60,
-	Hour = 1000 * 60 * 60,
-	Day = 1000 * 60 * 60 * 24,
-	Month = 1000 * 60 * 60 * 24 * (365 / 12),
-	Year = 1000 * 60 * 60 * 24 * 365
+	Minute = Second * 60,
+	Hour = Minute * 60,
+	Day = Hour * 24,
+	Week = Day * 7,
+	Month = Day * (365 / 12),
```
vladfrangu, line 12: Will I be pedantic about the fact a year isn't exactly 365 days, and a month isn't technically `365/12`? yes. 🤣

vladfrangu (changes requested):

packages/duration/src/lib/constants.ts
favna: Tbh we had that in the past too so lets not be pedantic lol

kyranet: I can do a semver-major change to make them `365.25` (and also normalize it everywhere), wdyt?

favna: I don't really want a semver major for that tbh

vladfrangu: What, it wouldn't be semver major to change it, I'd consider this a patch 🤔 . Can you explain why this would be major? 👁️

vladfrangu: That still doesn't answer my question! Yes, people rely on these values, but they also just use the enum access, and not the raw number 🤔 
It also only affects the biggest units, which most people don't use anyways, and I'd rather release a patch that have the *correct* values rather than avoid fixing this because its pseudo-major

favna: well.. fine.. so what will it be then? Either way, @kyranet can you fix that in #562 and change the commits from `refactor` to `fix`? This time you *are* going to have to rebase and use `edit` on each Kyra, sorry.

favna (approved):

favna (approved):

PR #567, opened 2023-03-17
feat: update repository to TypeScript 5.x

favna:
- feat(ts-config): update for TypeScript v5.x
- feat(ts-config): use `Node16` as default module
- fix(snowflake): ensure strings are never compared with bigints
- fix(result): add missing type modifier
- fix(discord.js-utilities): add missing type modifiers
- chore(*): bump dev dependency to TypeScript 5.x


packages/ts-config/tsconfig.json
```diff
@@ -5,15 +5,12 @@
...
-		"experimentalDecorators": true,
-		"forceConsistentCasingInFileNames": true,
 		"importHelpers": true,
-		"importsNotUsedAsValues": "error",
+		"verbatimModuleSyntax": true,
 		"incremental": true,
 		"lib": ["esnext"],
-		"module": "CommonJS",
+		"module": "Node16",
 		"moduleResolution": "Node",
```
vladfrangu, line 14: Should also be Node16!

vladfrangu (changes requested):

packages/ts-config/tsconfig.json
favna: No actually, setting both causes tonnes of these errors again (same as just setting `module` to `CommonJS`)

CleanShot 2023-03-17 at 22 29 45@2x

Note also how by default it is specifically `Node` when `module` is set to `node16`: https://www.typescriptlang.org/tsconfig#moduleResolution

vladfrangu: That sounds like a bug on TS side because I recall them saying you need both!

favna: Contrary, they dont. They don't even mention `moduleResolution` on the blog post where they announced `node16`: https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/.

Even if you include the typo'd `moduleresoluton`, that is still only 1 hit and doesn't talk about the new option for that configuration. 

Furthermore, the reference page for `moduleResolution` also doesn't list `node16`: https://www.typescriptlang.org/docs/handbook/module-resolution.html

favna:
@sapphiredev pack

favna:
@sapphiredev pack

c43721 (dismissed):

favna:
@sapphiredev pack

packages/snowflake/src/lib/Snowflake.ts
```diff
@@ -129,7 +129,8 @@ export class Snowflake {
 	public static compare(a: string | bigint, b: string | bigint): -1 | 0 | 1 {
 		if (typeof a === 'bigint' || typeof b === 'bigint') {
 			if (typeof a === 'string') a = BigInt(a);
-			else if (typeof b === 'string') b = BigInt(b);
+			if (typeof b === 'string') b = BigInt(b);
```
kyranet, line 132: This is unnecessary, because the check at 130 ensures that either `a` or `b` is `bigint`, so only one may be `string`. You're free to try to prove this with unit tests, although I think those exist already.

kyranet (changes requested):

packages/snowflake/src/lib/Snowflake.ts
favna: image

kyranet (approved):

vladfrangu (approved):

PR #580, opened 2023-04-16
feat(utilities): add `enumToObject`

### LINKED ISSUE #507: request: `@sapphire/bitfield` lacks support for 2-way enums

### Feature description

Depends on #502.

Implementing a `mapEnum` utility function to map TypeScript's 2-way enums into 1-way ones, as `BitField`'s constructor requires all values to be of the same type, and supporting them in the constructor can lead to much more complicated types, as opposed to a function, which would remove one step.

### Desired solution

The idea is to make a function that maps (both value and types) of 2-way enums (aka TypeScript enums) into 1-way ones, for example:

```typescript
enum State {
	On,
	Off
}
```

Results on the following:

```typescript
const State = {
	On: 0,
	Off: 1,
	0: 'On',
	1: 'Off'
};
```

The idea is to have a function, e.g. `mapEnum`, that maps the above object (and type) into:

```typescript
const State = {
	On: 0,
	Off: 1
};

type State = {
	On: 0;
	Off: 1;
};
```

So it can be passed into the BitField's constructor in the following manner:

```typescript
const PermissionsBitField = new BitField(mapEnum(PermissionsFlagsBits));
```

### Alternatives considered

Mapping in the constructor by ignoring values which type is `string` and are a property of `flags`, but then again types would need to handle this, and I fear they might be incredibly complex.

### Additional context

Conversation at https://github.com/sapphiredev/utilities/pull/502#discussion_r996082235

kyranet:
Fixes #507

### DIFF
--- a/packages/bitfield/src/lib/utilities.ts
+++ b/packages/bitfield/src/lib/utilities.ts
@@ -25,5 +25,5 @@ export function enumToObject<T extends object>(enumObject: T): { [K in Exclude<k
 		if (Number.isNaN(Number(key))) result[key] = value;
 	}
 
-	return result as any;
+	return result;
 }
--- a/packages/bitfield/src/lib/utilities.ts
+++ b/packages/bitfield/src/lib/utilities.ts
@@ -1,0 +1,29 @@
+/**
+ * Converts a TypeScript enum to a 1-way object, stripping out the number keys.
+ * @param enumObject The enum to convert
+ * @example
+ * ```typescript
+ * enum Permissions {
+ * 	Read: 1 << 0,
+ * 	Write: 1 << 1
+ * }
+ * // {
+ * // 	Read: 1,
+ * // 	Write: 2,
+ * // 	1: 'Read',
+ * // 	2: 'Write'
+ * // }
+ *
+ * enumToObject(Permissions);
+ * // { Read: 1, Write: 2 }
+ * ```
+ * @returns The mapped object
+ */
+export function enumToObject<T extends object>(enumObject: T): { [K in Exclude<keyof T, `${number}`>]: T[K] } {
+	const result = {} as any;
+	for (const [key, value] of Object.entries(enumObject)) {
+		if (Number.isNaN(Number(key))) result[key] = value;
+	}
+
+	return result as any;
+}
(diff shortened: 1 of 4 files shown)


packages/bitfield/src/lib/utilities.ts
```diff
@@ -0,0 +1,29 @@
...
+ * // 	1: 'Read',
+ * // 	2: 'Write'
+ * // }
+ *
+ * enumToObject(Permissions);
+ * // { Read: 1, Write: 2 }
+ * ```
+ * @returns The mapped object
+ */
+export function enumToObject<T extends object>(enumObject: T): { [K in Exclude<keyof T, `${number}`>]: T[K] } {
```
vladfrangu, line 22: Doesn't this allow passing in null too?

kyranet: No, we removed `NonNullObject` because of this.

packages/bitfield/src/lib/utilities.ts
```diff
@@ -0,0 +1,29 @@
...
+	const result = {} as any;
+	for (const [key, value] of Object.entries(enumObject)) {
+		if (Number.isNaN(Number(key))) result[key] = value;
+	}
+
+	return result as any;
```
favna, line 28: any? Use Generic type.

favna (changes requested):

packages/bitfield/src/lib/utilities.ts
```diff
@@ -0,0 +1,29 @@
...
+ * // 	2: 'Write'
+ * // }
+ *
+ * enumToObject(Permissions);
+ * // { Read: 1, Write: 2 }
+ * ```
+ * @returns The mapped object
+ */
+export function enumToObject<T extends object>(enumObject: T): { [K in Exclude<keyof T, `${number}`>]: T[K] } {
+	const result = {} as any;
```
favna, line 23: There's still an any here...

favna (changes requested):

favna (approved):
Loads better. Now if anything ever changes in TS that would affect this we'll actually know.

PR #584, opened 2023-04-24
feat(PaginatedMessage): add method to update pages after response

killbasa:
Adds a method to update page content after `PaginatedMessage` has run, allowing more interactivity when handling buttons.

Example:
```ts
run: async ({ handler }) => {
    const value = Math.floor(Math.random() * 10);

    await handler.updateCurrentPage({
        embeds: [
            new EmbedBuilder().setDescription(`${value}`)
        ]
    });
}
```


favna:
@sapphiredev pack

PR #585, opened 2023-04-25
feat(discord.js-utilities): add `JITPaginatedMessage` for per-page action support

killbasa:
`JITPaginatedMessage` adds support for per-page actions.

`PaginatedMessage` functionality has not changed for end-users. Internally, `actions` are now resolved to `components` on each page change and cached.

**TODO**:
- [ ] Test all `PaginatedMessage` methods with the new action loading
- [ ] Improve DX for adding/setting pages to `JITPaginatedMessage`
- [ ] See if `components` can be cached on first run instead of just page changes


killbasa:
Closed in favor of #587

PR #590, opened 2023-05-05
chore(deps): update `vitest`

kyranet:
The methods were renamed around, the two methods we used were clearing the implementation, `mockClear` only resets the state.

https://github.com/vitest-dev/vitest/blob/77d071a91f81b6faef28cb30c6c5b617a661459d/packages/spy/src/index.ts#L224-L242


favna (approved):
tyvm kyra <3

PR #594, opened 2023-05-09
refactor(bitfield): call entries only once

kyranet:
This drastically improves performance of the `toArray` and `toObject` methods.

For the record about "but we don't use symbols for privates anywhere!", we still support Node.js v16, which has horrible `#private` performance, and tsup compiles the `#private` fields as weakmaps, nullifying the performance gains.

### DIFF
--- a/packages/bitfield/src/lib/BitField.ts
+++ b/packages/bitfield/src/lib/BitField.ts
@@ -1,4 +1,4 @@
-const FlagEntriesSymbol = Symbol('@sapphire/bitfield:bitfield.flags.entries');
+const FlagEntriesSymbol = Symbol('@sapphire/bitfield.flags.entries');
 
 export class BitField<Flags extends Record<string, number> | Record<string, bigint>> {
 	public readonly type: Flags[keyof Flags] extends number ? 'number' : 'bigint';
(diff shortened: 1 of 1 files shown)


packages/bitfield/src/lib/BitField.ts
```diff
@@ -1,8 +1,11 @@
+const FlagEntriesSymbol = Symbol('@sapphire/bitfield:bitfield.flags.entries');
```
vladfrangu, line 1: no need to complicate this so much since its Symbol, not Symbol.for

```suggestion
const FlagEntriesSymbol = Symbol('@sapphire/bitfield.flags.entries');
```

vladfrangu (changes requested):

vladfrangu (dismissed):

packages/bitfield/src/lib/BitField.ts
```diff
@@ -1,8 +1,11 @@
+const FlagEntriesSymbol = Symbol('@sapphire/bitfield.flags.entries');
```
favna, line 1: Use `#private` fields instead of symbols, also see other PR.

favna (changes requested):

favna (approved):

PR #595, opened 2023-05-09
feat(snowflake): expose `processId` and `workerId`

kyranet:
This PR also switches `#private` properties to symbol properties for higher performance. Also added missing tests to near 100% coverage.


vladfrangu (dismissed):

favna (changes requested):
#private properties performance has been fixed like 6 months ago or more in Node 18. So that can be reverted.

PR #596, opened 2023-05-11
refactor(snowflake): handle out-of-bounds `increment` correctly

kyranet:
This fixes several issues at once, and also increases performance.

- The code was cycling at 4095, making the effective range [0, 4094] instead of [0, 4095].
- When passing a custom increment over the limit (>4095), it would be set to 0, which can cause duplication issues.
- When passing a custom increment that was <=4095, it would call the `else` branch which would then override with the defaults.
- Were custom increments to be allowed, negative values would overflow the rest of the data. For example, fixing the branching code:
  ```js
  DiscordSnowflake.generate({ increment: -1n, timestamp: 1600000000000n, processId: 1n, workerId: 0n });
  ```
  Would return... `-1`, which is can be deconstructed as:
  ```js
  {
    id: -1n,
    timestamp: 1420070399999n,
    workerId: 31n,
    processId: 31n,
    increment: 4095n,
    epoch: 1420070400000n
  }
  ```
  Which is nowhere close to the input data.

Also, the new code should be a bit faster now that it accesses the private field exactly only twice, rather than 2 or 3 times (when cycling) and uses less conditional branches.

This PR also adds exports for the maximum values each field can have: `MaximumWorkerId`, `MaximumProcessId`, and `MaximumIncrement`.

### DIFF
--- a/packages/snowflake/src/lib/Snowflake.ts
+++ b/packages/snowflake/src/lib/Snowflake.ts
@@ -131,7 +131,6 @@ export class Snowflake {
 		}
 
 		if (typeof increment !== 'bigint') {
-			// The increment was within
 			increment = this[IncrementSymbol];
 			this[IncrementSymbol] = (increment + 1n) & MaximumIncrement;
 		}
--- a/packages/snowflake/src/lib/Snowflake.ts
+++ b/packages/snowflake/src/lib/Snowflake.ts
@@ -4,6 +4,21 @@ const ProcessIdSymbol = Symbol('@sapphire/snowflake.processId');
 const WorkerIdSymbol = Symbol('@sapphire/snowflake.workerId');
 
 /**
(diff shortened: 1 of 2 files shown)


kyranet:
I have no idea if this is https://github.com/sapphiredev/utilities/labels/semver%3Aminor (because the new exports) or https://github.com/sapphiredev/utilities/labels/semver%3Amajor (because the change in behaviour when passing out-of-bounds increments).

Regardless, the method was incredibly broken to the point those behaviour changes get into a grey area where they were simply impossible due to implementation issues.

vladfrangu:
It is definitely not https://github.com/sapphiredev/utilities/labels/semver%3Amajor, as this is a bug fix!

vladfrangu (approved):

PR #601, opened 2023-05-22
refactor(PaginatedMessage): update types and make `run` on actions optional

killbasa:
There were a few types in the `PaginatedMessage` class which didn't reflect how the class actually resolves things. This brings them in-line with their actual behaviour.

## Changes:

- Allowing `run` on actions to be optional means they can be resolved in interaction handlers rather than just `PaginatedMessage`. This is useful if the button/menu contains a lot of business logic and you want it to have its own piece, or you're passing data in the custom ID.

- `PaginatedMessageMessageOptionsUnion` now omits the `components` property since it was never handled in the first place. Originally, you could pass `components` in the message options without any issues, however the class never used that property and instead resolved the components from `actions`.

- The callback union type is removed from `messages`. This is due to pages _always_ being processed with `resolvePage`, meaning they can never be a callback. The only reason this was on `messages` on the first place was because `pages` from the constructor were immediately placed in the `messages` array without any processing.

    https://github.com/sapphiredev/utilities/blob/db130684ff3c1680f736451361eec80686a783e0/packages/discord.js-utilities/src/lib/PaginatedMessages/PaginatedMessage.ts#L240-L244

- Added utility functions for type-guarding actions to make editing actions for `updateCurrentPage` easier.

- Cleaned up some `as` usages.

## Breaking changes:

- `updateCurrentPage` components are now resolved from the `actions` property instead of the `components` property. This makes it consistent with all the other page manipulation methods (ie `addPage`, `addPages`, pages provided through the constructor, etc).

    <details>

    <summary>Example: updating a string menu's options</summary>

    ```ts
    actions: [
        {
            type: ComponentType.StringSelect,
            customId: 'custom-id-1',
            options: [
                { label: 'option-1', value: 'value-1' },
                { label: 'option-2', value: 'value-2' },
                { label: 'option-3', value: 'value-3' }
            ]
        },
        {
            type: ComponentType.Button,
            customId: 'custom-id-2',
            style: ButtonStyle.Primary,
            label: 'Button',
            run: async ({ handler }): Promise<void> => {
                const value = Math.floor(Math.random() * 10);

                const page = await handler.getPageOptions(handler.index);

                const action = page?.actions?.at(0);
                if (action && isActionStringMenu(action)) {
                    action.options = [
                        { label: `option-1-${value}`, value: 'value-1' },
                        { label: `option-2-${value}`, value: 'value-2' },
                        { label: `option-3-${value}`, value: 'value-3' }
                    ];
                }

                await handler.updateCurrentPage({
                    ...page,
                    actions: page?.actions ? page.actions : []
                });
            }
        }
    ]
    ```
    </details>

- Removed `getComponents` since it is effectively useless due to the above change.

### DIFF
--- a/packages/discord.js-utilities/README.md
+++ b/packages/discord.js-utilities/README.md
@@ -19,7 +19,7 @@ There many functions that are always extremely useful to have when developing bo
 
 ## Features
 
--   Written in TypeScript
+-   Written in TypeScript.
 -   Bundled with esbuild to provide size-optimal bundles
 -   Offers CommonJS and ESM bundles
 
--- a/packages/discord.js-utilities/README.md
+++ b/packages/discord.js-utilities/README.md
@@ -62,4 +62,4 @@ Thank you to all the people who already contributed to Sapphire!
 
 [contributing]: https://github.com/sapphiredev/.github/blob/main/.github/CONTRIBUTING.md
 [@sapphire/discord-utilities]: https://www.npmjs.com/package/@sapphire/discord-utilities
-[discord.js]: https://discord.js.org
+[discord.js]: https://discordjs.dev
--- a/packages/discord.js-utilities/README.md
+++ b/packages/discord.js-utilities/README.md
@@ -20,7 +20,7 @@ There many functions that are always extremely useful to have when developing bo
 ## Features
 
 -   Written in TypeScript.
--   Bundled with esbuild to provide size-optimal bundles
+-   Bundled with esbuild to provide size-optimal bundles.
 -   Offers CommonJS and ESM bundles
 
 ## Installation
--- a/packages/discord.js-utilities/README.md
+++ b/packages/discord.js-utilities/README.md
@@ -21,7 +21,7 @@ There many functions that are always extremely useful to have when developing bo
 
 -   Written in TypeScript.
 -   Bundled with esbuild to provide size-optimal bundles.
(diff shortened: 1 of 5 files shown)


vladfrangu (dismissed):
Looks _fine_ but pls test it and make sure it doesn't break something random

favna:
Code looks good, going to test now

favna:
@sapphiredev pack

favna:
Dumping this here as a backup but this is what I want git-cliff to pick up. Now to figure out how to make it so.
```
refactor(PaginatedMessage): allow PaginatedMessage actions `run` to be optional
and instead get handled by Sapphire interaction-handlers
refactor(PaginatedMessage): `PaginatedMessageMessageOptionsUnion` now omits `components`
as it was not handled anyway.
refactor(PaginatedMessage): The callback union is removed from `messages` property.
pages are *always* processed with `resolvePage` meaning they can never be a callback.
feat(PaginatedMessage): Added utility functions for type-guarding action variants.
refactor(PaginatedMessage): Cleaned up some internal type casting.

BREAKING CHANGE: `updateCurrentPage` components are now resolved from the `actions` property
instead of the `components` property. This makes it consistent with all the other
page manipulation methods

BREAKING CHANGE: Removed `getComponents` since it is effectively useless due to the above change.
```

favna (approved):

PR #604, opened 2023-06-07
feat(utilities): add retry function

favna:
Thanks goes to @kyranet


vladfrangu (dismissed):

packages/utilities/src/lib/retry.ts
```diff
@@ -0,0 +1,23 @@
+import type { Awaitable } from './types';
+
+/**
+ * This code is used to retry a function call a specified number of times.
+ * The function will retry the call to the callback function until it either succeeds or it runs out of retries.
+ * @param  cb The function to be retried is passed in as a callback function.
+ * @param  retries The number of retries is also passed in as a parameter.
```
kyranet, line 7:
```suggestion
 * @param cb The function to be retried is passed in as a callback function.
 * @param retries The number of retries is also passed in as a parameter.
```

packages/utilities/src/lib/retry.ts
```diff
@@ -0,0 +1,23 @@
+import type { Awaitable } from './types';
+
+/**
+ * This code is used to retry a function call a specified number of times.
+ * The function will retry the call to the callback function until it either succeeds or it runs out of retries.
```
kyranet, line 5:
```suggestion
 * Calls the callback function until it either succeeds or it runs out of retries.
```

packages/utilities/tests/retry.test.ts
```diff
@@ -0,0 +1,26 @@
+import { retry } from '../src/lib/retry';
+
+describe('retry', () => {
+	test('GIVEN a simple string return THEN returns the same on first attempt', async () => {
+		const result = await retry(() => 'test', 3);
```
kyranet, line 5: If we use `vi.fn`, we can test how many times the function is called to ensure the amount is correct.

packages/utilities/tests/retry.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+		expect(result).toBe('test');
+	});
+
+	test('GIVEN a twice failing function THEN returns the third result', async () => {
+		let counter = 0;
+		const cb = () => {
```
kyranet, line 12: Ditto

kyranet (changes requested):
I wonder if we should have a `retrySync` (or a sync `retry` with `retryAsync`). 🤔

kyranet (approved):

vladfrangu (approved):

kyranet (dismissed):

kyranet (approved):

vladfrangu (changes requested):
Forgot to re-export in index.ts :megu:

vladfrangu (approved):

PR #616, opened 2023-07-07
feat(utilities): add poll function

favna:


packages/utilities/src/lib/poll.ts
```diff
@@ -0,0 +1,48 @@
+import { sleep } from './sleep';
+
+/**
+ * Executes a function {@link cb} and validates the result with function {@link cbCondition},
+ * and repeats this until {@link cbCondition} returns `true` or the {@link timeoutMilliseconds} is reached.
+ *
+ * For a synchronous variant, see [pollSync](./pollSync.d.ts).
```
vladfrangu, line 7: Thats a wacky way to link it

packages/utilities/src/lib/poll.ts
```diff
@@ -0,0 +1,48 @@
...
+ * @param cbCondition A function that when given the result of `fn` should return `true` if the polling should stop and should return `false` if the polling should continue.
+ * @param intervalMilliseconds The interval in between each poll.
+ * @param timeoutMilliseconds The maximum amount of time to poll before timing out and throwing an error.
+ * @param verbose Whether to log to the console on each polling interval, allowing the tracing of the amount of required attempts.
+ * @returns The result of {@link cb} as soon as {@link cbCondition} returns `true`, or an error if {@link timeoutMilliseconds} is reached.
+ * @throws If {@link timeoutMilliseconds} is reached.
+ */
+export async function poll<T>(
+	cb: () => T,
+	cbCondition: (arg: T) => boolean,
```
vladfrangu, line 18: Might as well make these methods return Awaitables

vladfrangu (changes requested):

packages/utilities/src/lib/poll.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @param cbCondition A function that when given the result of `fn` should return `true` if the polling should stop and should return `false` if the polling should continue.
+ * @param intervalMilliseconds The interval in between each poll.
+ * @param timeoutMilliseconds The maximum amount of time to poll before timing out and throwing an error.
+ * @param verbose Whether to log to the console on each polling interval, allowing the tracing of the amount of required attempts.
+ * @returns The result of {@link cb} as soon as {@link cbCondition} returns `true`, or an error if {@link timeoutMilliseconds} is reached.
+ * @throws If {@link timeoutMilliseconds} is reached.
+ */
+export async function poll<T>(
+	cb: () => Awaitable<T>,
+	cbCondition: (arg: Awaitable<T>) => boolean,
```
kyranet, line 19: You want the unwrapped `Promise<T>`:
```suggestion
	cbCondition: (arg: Awaited<T>) => boolean,
```

packages/utilities/src/lib/poll.ts
```diff
@@ -0,0 +1,49 @@
...
+	intervalMilliseconds: number,
+	timeoutMilliseconds: number,
```
kyranet, line 21: I'm not convinced on the `Milliseconds` suffix, furthermore, I'm also not convinced on `timeoutMilliseconds` being a number rather than an `AbortSignal` which lines up better.

It would also be optional, so you can make an object option with `verbose`.

packages/utilities/src/lib/poll.ts
```diff
@@ -0,0 +1,49 @@
...
+
+	const startTime = Date.now();
+	let result = await cb();
+
+	while (!cbCondition(result)) {
+		if (verbose) {
+			console.log(`Waiting ${intervalMilliseconds} ms before polling again...`);
+		}
+
+		await sleep(intervalMilliseconds);
```
kyranet, line 38: Following the previous suggestion, supporting an `AbortSignal` would allow to make this `sleep` call abortable.

Furthermore, `startTime + intervalMilliseconds` may be `> timeoutMilliseconds` even without the `await cb();` call (which is a lot more likely than a timeout from `cb()`'s side). If we use an `AbortSignal`, it would throw instead, which simplifies a lot of logic in this function.

Also, we should support [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) or something, to avoid overloading the resources on `cb()`'s side, it's best practices after all for some resources such as querying a website that's currently not available.

packages/utilities/src/lib/poll.ts
```diff
@@ -0,0 +1,49 @@
...
+	verbose?: boolean
+): Promise<Awaitable<T>> {
+	if (timeoutMilliseconds < 0) {
+		throw new RangeError('Expected timeoutMilliseconds to be a number >= 0');
+	}
+
+	if (timeoutMilliseconds === 0) return cb();
+
+	const startTime = Date.now();
+	let result = await cb();
```
kyranet, line 31: Following the previous suggestion, supporting an `AbortSignal` would allow to make this call abortable, e.g. when the code it calls supports signals.

packages/utilities/src/lib/poll.ts
```diff
@@ -0,0 +1,49 @@
...
+		}
+
+		await sleep(intervalMilliseconds);
+
+		result = await cb();
+
+		const currentTime = Date.now();
+		if (currentTime - startTime > timeoutMilliseconds) {
+			throw new Error(`Polling task timed out after ${timeoutMilliseconds} milliseconds`);
+		}
```
kyranet, line 45: I'm concerned that without an `AbortSignal`, `cb()`'s execution is finished and fully processed only for an error to be thrown instead of being able to recover the result somehow.

With an `AbortSignal`, resources would be freed and we'd have a way to stop processing that would otherwise be useless.

kyranet (changes requested):
I'm sorry I'm basically telling you to rewrite almost the entire `poll` function, but not only the implementation looks dated, it also doesn't align with the common design on async APIs in JavaScript, which is to make them abortable.

packages/utilities/src/lib/poll.ts
```diff
@@ -0,0 +1,76 @@
...
+	if (!Number.isSafeInteger(waitBetweenRetries) || waitBetweenRetries < 0) {
+		throw new RangeError('Expected waitBetweenRetries to be a positive safe integer');
+	}
+
+	signal?.throwIfAborted();
+	let result = await cb(signal);
+	if (maximumRetries === 0) return result;
+
+	let retries = 0;
+	while (retries < maximumRetries && !cbCondition(result, signal)) {
```
vladfrangu, line 63: I've said it before, cbCondition should support promises too 👀

kyranet (approved):

packages/utilities/tests/poll.test.ts
```diff
@@ -0,0 +1,196 @@
...
+		test('GIVEN verbose but no waitBetweenRetries THEN does not call console.log', async () => {
+			const cb = vi.fn<[], string>().mockReturnValueOnce(fail).mockReturnValueOnce(pass);
+			const consoleLog = vi.spyOn(console, 'log').mockImplementation(() => undefined);
+			const result = poll(cb, cbConditionRaw, { verbose: true });
+
+			await expect(result).resolves.toBe(pass);
+			expect(consoleLog).toHaveBeenCalledTimes(0);
+		});
+
+		test('GIVEN verbose but no waitBetweenRetries THEN does not call console.log', async () => {
```
kyranet, line 186:
```suggestion
		test('GIVEN verbose and waitBetweenRetries THEN calls console.log on retry', async () => {
```

packages/utilities/tests/pollSync.test.ts
```diff
@@ -0,0 +1,194 @@
...
+		test('GIVEN verbose but no waitBetweenRetries THEN does not call console.log', () => {
+			const cb = vi.fn<[], string>().mockReturnValueOnce(fail).mockReturnValueOnce(pass);
+			const consoleLog = vi.spyOn(console, 'log').mockImplementation(() => undefined);
+			const result = pollSync(cb, cbConditionRaw, { verbose: true });
+
+			expect(result).toBe(pass);
+			expect(consoleLog).toHaveBeenCalledTimes(0);
+		});
+
+		test('GIVEN verbose but no waitBetweenRetries THEN does not call console.log', () => {
```
kyranet, line 184:
```suggestion
		test('GIVEN verbose and waitBetweenRetries THEN calls console.log on retry', () => {
```

kyranet (changes requested):

kyranet (approved):

PR #634, opened 2023-08-13
Use await in case of asynchronous method or fallback.

notsapinho:
In case the method or fallback is async, not using `await` will result the return of Promise resulting in a false value making the function not work.

### DIFF
--- a/packages/decorators/src/utils.ts
+++ b/packages/decorators/src/utils.ts
@@ -78,7 +78,7 @@ export function createFunctionPrecondition(precondition: FunctionPrecondition, f
 
 		descriptor.value = async function descriptorValue(this: (...args: any[]) => any, ...args: any[]) {
 			const canRun = await precondition(...args);
-			return canRun ? method.call(this, ...args) : fallback.call(this, ...args);
+			return canRun ? await method.call(this, ...args) : await fallback.call(this, ...args);
 		} as unknown as undefined;
 	});
 }


kyranet:
This won't fix it, `return await` is identical to `return` if it's not within a `try...catch/finally`. Promises flatten themselves so a promise of a promise acts like a flat promise.

notsapinho:
> This won't fix it, `return await` is identical to `return` if it's not within a `try...catch/finally`. Promises flatten themselves so a promise of a promise acts like a flat promise.

As I explained on Sapphire's Discord, it fixes it, since the chatInputRun is wrapped inside a [Result.fromAsync](https://github.com/sapphiredev/plugins/blob/79d7230c84ad59338197d8109e6f3ba46a5c2158/packages/subcommands/src/lib/Subcommand.ts#L303C3-L303C3). The [decorator](https://github.com/sapphiredev/utilities/blob/141fc267cb3ad4ff1ddbf5d05fd1478da68d7476/packages/decorators/src/utils.ts#L73) changes the chatInputRun from the class, just test it and you will see that it makes differential.

favna:
@notsapinho can you provide a github repository where the original issue can be reproduced?

notsapinho:
> @notsapinho can you provide a github repository where the original issue can be reproduced?

I can't give you out the repo because is private, but the decorator that caused the problem is this one:

```ts
import { UserError } from "@sapphire/framework";
import type { Command } from "@kaname-png/plugin-subcommands-advanced";

import { PermissionsBitField, type PermissionResolvable } from "@lib/bitfields";
import { ErrorIdentifiers } from "@lib/utils";
import { checkPermissionForMember, createFunctionPrecondition, type CheckPermissionOptions } from "@lib/utils/helpers";

export const HasPermission = (
    permissionResolvable: PermissionResolvable,
    opts: CheckPermissionOptions = {
        checkRoles: true,
        adminOverride: true
    }
): MethodDecorator => {
    return createFunctionPrecondition(async (interaction: Command.ChatInputInteraction<"cached">) => {
        const resolved = new PermissionsBitField(permissionResolvable);
        const has = await checkPermissionForMember(interaction.member!, resolved, opts);

        if (!has)
            throw new UserError({
                identifier: ErrorIdentifiers.MissingPermissions,
                message: "> Você não possui permissões suficientes para executar este comando."
            });

        return true;
    });
};
```

  _The `createFunctionPrecondition` is the modified version that I proposed on the pull request, using the default one causes the command or whatever interaction not work as I explained earlier._

favna:
The sample above is missing the impl of all of `@lib/utils/helpers` which means I cannot use it to reproduce this issue. Furthermore, I'm not asking you to share the repository where you have this issue, I'm asking for a repository where the issue can be reproduced. 

Resources:
- https://minimum-reproduction.wtf
- https://antfu.me/posts/why-reproductions-are-required
- https://apache.googlesource.com/cordova-contribute/+/HEAD/create-reproduction.md
- https://gist.github.com/Rich-Harris/88c5fc2ac6dc941b22e7996af05d70ff

notsapinho:
> The sample above is missing the impl of all of `@lib/utils/helpers` which means I cannot use it to reproduce this issue. Furthermore, I'm not asking you to share the repository where you have this issue, I'm asking for a repository where the issue can be reproduced.
> 
> Resources:
> 
>     * https://minimum-reproduction.wtf
> 
>     * https://antfu.me/posts/why-reproductions-are-required
> 
>     * https://apache.googlesource.com/cordova-contribute/+/HEAD/create-reproduction.md
> 
>     * https://gist.github.com/Rich-Harris/88c5fc2ac6dc941b22e7996af05d70ff

Here is the [repo](https://github.com/notsapinho/sapphire-decorator-problem). To test it, use the "/test" command. First, try running it without the 'ban' permission, and then try again with the 'ban' permission.

favna:
With all the latest dependencies (see below) and an **unmodified** @sapphire/decorators I cannot reproduce this issue with the repo above.

Running the command simply gives me: `(2) HAS: true`. 

That said, because this was originally raised as an issue when using permission checking decorators for subcommands I am going to close this as "wont' fix", because as of more recent versions of @sapphire/plugin-subcommands there is now native support for [Sapphire preconditions](https://www.sapphirejs.dev/docs/Guide/preconditions/what-are-preconditions) for subcommands and so the decorator wouldn't even be needed at all anymore. Furthermore, I also agree with Kyra that `return` and `return await` are identical and my own testing also proves this.

<details><summary>expand for dependencies</summary>

```json
    "dependencies": {
        "@sapphire/decorators": "^6.0.2",
        "@sapphire/framework": "^4.7.2",
        "@sapphire/plugin-logger": "^3.0.6",
        "@skyra/env-utilities": "^1.2.1",
        "discord.js": "^14.14.1"
    },
    "devDependencies": {
        "@sapphire/ts-config": "^5.0.0",
        "@swc/cli": "^0.1.63",
        "@swc/core": "^1.3.96",
        "@types/node": "^20.9.0",
        "chokidar": "^3.5.3",
        "prettier": "^3.1.0",
        "typescript": "^5.2.2"
    },
```

</details>

PR #637, opened 2023-08-24
Update the tsconfig to better support TypeScript 5.2

favna:
> **Warning**
> needs rebase & merge to ensure that the breaking change doesn't show up for the other packages

---

As TS 5.2 is really pushing for separating CJS and ESM entirely I have decided to completely restructure our config. This is a majorly breaking change and an upgrading guide is provided. For review you really only need to look at commit [a5ba402](https://github.com/sapphiredev/utilities/commit/a5ba402305ddcbb8dc97e4474c4aa8586c9a9810), the other commit is just updating the packages in this repo itself.

Closes #576


packages/ts-config/src/cjs/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+	"compilerOptions": {
+		"module": "CommonJS",
```
vladfrangu, line 3: nononononononononono, we should NOT be doing this. It should be the same as ESM, otherwise TS will not know about exports stuff.

vladfrangu (changes requested):

vladfrangu (approved):

PR #662, opened 2023-11-10
docs(discord.js-utilities): update discord.js.org links

kyranet:
Now that the rewrite is up and live, I updated the links. `@discord.js-utilities` was the only package referencing djs's docs, so only one package was updated.


vladfrangu (approved):

favna (approved):
Oh I can finally archive favna/discord.js-docs-parser then because Souji will already or will soon migrate djs-helper bot

PR #663, opened 2023-11-10
feat(bit-array): initial commit

kyranet:
This is a dependency for `@sapphire/string-store`'s unaligned (and more compact) data storage, based on [`TypedArray`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)'s API, but using bits instead of bytes.

Pending to add the missing code, and **_a lot_** of tests.


favna:
Can't you just make this part of the code of string-store. Why does this need to be its own package. This way it just adds more unnecessary maintenance overhead.

PR #665, opened 2023-11-14
fix(discord.js-utilities): fixed `isNsfwChannel` typeguard by including `ChannelType.GuildMedia`

favna:


packages/discord.js-utilities/src/lib/type-guards.ts
```diff
@@ -211,6 +213,7 @@ export function isNsfwChannel(channel: ChannelTypes | Nullish): boolean {
 		case ChannelType.AnnouncementThread:
 		case ChannelType.PrivateThread:
 		case ChannelType.PublicThread:
+		case ChannelType.GuildMedia:
```
vladfrangu, line 216: This is like a forum channel

vladfrangu (approved):

PR #694, opened 2023-12-18
fix(tsconfig): add DOM types to lib

favna:
Too often does it happen that @types/node does not expose everything that is officially available in the Global scope to the Global scope. A perfect example of this is that without DOM types when compiling a project that uses @sapphire/fetch you get the error:
```
node_modules/@sapphire/fetch/dist/cjs/index.d.ts:264:12 - error TS2304: Cannot find name 'BodyInit'.

264     body?: BodyInit | Record<any, any>;
               ~~~~~~~~


Found 1 error in node_modules/@sapphire/fetch/dist/cjs/index.d.ts:264
```

And this is just one possible example among others. As Node and browsers inch ever close to one another in terms of API this problem is only going to increase I think.

### DIFF
--- a/packages/ts-config/src/tsconfig.json
+++ b/packages/ts-config/src/tsconfig.json
@@ -8,7 +8,7 @@
 		"esModuleInterop": true,
 		"importHelpers": false,
 		"incremental": true,
-		"lib": ["esnext"],
+		"lib": ["ESNext", "DOM"],
 		"module": "Node16",
 		"moduleResolution": "Node16",
 		"newLine": "lf",


vladfrangu:
Fun, node 20 has these globally defined, but node 18 doesn't?

packages/ts-config/src/tsconfig.json
```diff
@@ -8,7 +8,7 @@
 		"esModuleInterop": true,
 		"importHelpers": false,
 		"incremental": true,
-		"lib": ["esnext"],
+		"lib": ["ESNext", "DOM"],
```
vladfrangu, line 11: God to the no, if you want to add the types, add it just to that one file, not globally across all projects.

vladfrangu (changes requested):

favna:
Closing this in favour of #695

PR #700, opened 2023-12-28
feat(utilities): add omitKeysFromObject utility function

favna:
Ported from https://github.com/skyra-project/skyra/pull/2505#discussion_r1390246482 where I commented that the new utility function from Skyra would do well to be ported to Sapphire Utilities.

### DIFF
--- a/packages/utilities/src/lib/omitKeysFromObject.ts
+++ b/packages/utilities/src/lib/omitKeysFromObject.ts
@@ -10,7 +10,7 @@ import { deepClone } from './deepClone';
  * @param keys - The keys to omit from the object.
  * @returns A new object without the specified keys.
  */
-export function omitKeysFromObject<TO extends object, KTO extends keyof TO>(source: TO, ...keys: readonly KTO[]): Omit<TO, KTO> {
+export function omitKeysFromObject<Object extends object, ObjectKeys extends keyof Object>(source: Object, ...keys: readonly ObjectKeys[]): Omit<Object, ObjectKeys> {
 	const clone = deepClone(source);
 
 	for (const key of keys) {
--- a/packages/utilities/src/lib/omitKeysFromObject.ts
+++ b/packages/utilities/src/lib/omitKeysFromObject.ts
@@ -10,7 +10,10 @@ import { deepClone } from './deepClone';
  * @param keys - The keys to omit from the object.
  * @returns A new object without the specified keys.
  */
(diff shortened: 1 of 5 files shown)


vladfrangu (dismissed):

packages/utilities/src/lib/omitKeysFromObject.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Clones the source object using {@link deepClone} then deletes the specified keys with {@link Reflect.deleteProperty}
+ *
+ * @template TO - The object type.
+ * @template KTO - The keys of the object type.
+ *
+ * @param source - The input object.
+ * @param keys - The keys to omit from the object.
+ * @returns A new object without the specified keys.
+ */
+export function omitKeysFromObject<TO extends object, KTO extends keyof TO>(source: TO, ...keys: readonly KTO[]): Omit<TO, KTO> {
```
kyranet, line 13: Extending the names would be more readable:
```suggestion
export function omitKeysFromObject<Object extends object, ObjectKeys extends keyof Object>(source: Object, ...keys: readonly ObjectKeys[]): Omit<Object, ObjectKeys> {
```

packages/utilities/tests/omitKeysFromObject.test.ts
```diff
@@ -0,0 +1,43 @@
+import { omitKeysFromObject } from '../src/lib/omitKeysFromObject';
+
+describe('omitKeysFromObject', () => {
```
kyranet, line 3: There should be a test checking that `source` and `result` are different objects, and that `source` isn't mutated.

kyranet (changes requested):

kyranet (dismissed):

kyranet (approved):

packages/utilities/src/lib/omitKeysFromObject.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Clones the source object using {@link deepClone} then deletes the specified keys with {@link Reflect.deleteProperty}
+ *
+ * @template TO - The object type.
+ * @template KTO - The keys of the object type.
+ *
+ * @param source - The input object.
+ * @param keys - The keys to omit from the object.
+ * @returns A new object without the specified keys.
+ */
+export function omitKeysFromObject<Object extends object, ObjectKeys extends keyof Object>(
```
vladfrangu, line 13: @kyranet shouldn't we be using {} instead of object?

vladfrangu (approved):

packages/utilities/src/lib/omitKeysFromObject.ts
kyranet: No, recent TS versions made `object` exclude `null`.

PR #724, opened 2024-03-11
chore(event-iterator): use `ts-ignore` to silence compiler issue

yuansheng1549:
This pr fixes `event-iterator` unused `ts-expect-error` directive.


vladfrangu (approved):
Could've also just removed the ts-expect-error, but this works too

yuansheng1549:
> Could've also just removed the ts-expect-error, but this works too

When removing the ts-expect-error, it will failed to build

image

favna (approved):

PR #728, opened 2024-03-25
docs(eslint-config): add note on how to use with flat config

HigherOrderLogic:
Add instruction on how to use with Esline flat config.

If you guys want, I can include the code into the package, eg: `@sapphire/eslint-config/flat`.

### DIFF
--- a/packages/eslint-config/README.md
+++ b/packages/eslint-config/README.md
@@ -44,6 +44,22 @@ Or to `eslintrc.js` / `.eslintrc.json`:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


favna:
We've been thinking about making @sapphire/eslint-config an opinionated extension on top of https://github.com/iCrawl/eslint-config-neon, once we finally get around doing that we'll also add support for flat config as Neon already has it.

HigherOrderLogic:
So should I close this PR for now?

favna (approved):
Nah I guess we can keep this

PR #737, opened 2024-04-26
feat: add `iterator-utilities` package

kyranet:
### 🛬 TODO

- [x] Write shared function for resolving iterators/iterables
- [x] Write tests
- [x] Make docs more consistent and add examples
- [x] Write README.md
- [x] Tweak the `subpath-updater.cjs` script to filter out the files from `iterator-utilities/src/lib/common` (cc: @favna)


vladfrangu (dismissed):

packages/iterator-utilities/src/lib/dropWhile.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+export function dropWhile<const ElementType, const FilteredType extends ElementType>(
+	iterable: IterableResolvable<ElementType>,
+	callbackFn: (element: ElementType, index: number) => element is FilteredType
+): IterableIterator<Exclude<ElementType, FilteredType>>;
+export function dropWhile<const ElementType>(
+	iterable: IterableResolvable<ElementType>,
+	callbackFn: (element: ElementType, index: number) => boolean
+): IterableIterator<ElementType>;
+
```
favna, line 31: Why the empty line here?

packages/iterator-utilities/src/lib/except.ts
favna, line 1: Shouldn't this be called `difference`? 

> Q: *how is the process called where you have 2 lists and you extract those elements which exist only in the first one*
>
>
> A: The process you’re describing, where you extract elements that exist only in the first list, is commonly referred to as finding the symmetric difference between two lists. Let me explain how this works:

> **Symmetric Difference:**
> 
> Given two sets (or lists), the symmetric difference contains elements that are present in either of the sets but not in their intersection.
> 
> Mathematically, if we have sets A and B, the symmetric difference is denoted as (A \Delta B).
> 
> In the context of your question, it means finding elements that are unique to either the first list or the second list (but not both).


Alternatively maybe `omit` because that's also how TS calls it for the `Omit<A, B>` type.

packages/iterator-utilities/src/lib/intersect.ts
favna, line 1: Add a remark that this is the opposite of `except`, or rather `difference` / `omit` after the rename.

packages/iterator-utilities/src/lib/last.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @remarks
+ *
+ * This function consumes the entire iterator to find the last value.
+ */
+export function last<const ElementType>(iterable: IterableResolvable<ElementType>): ElementType | undefined {
+	let last: ElementType | undefined;
+	for (const value of toIterableIterator(iterable)) {
+		last = value;
+	}
```
favna, line 27: Wouldn't it be better to do this with a while loop where you call `.next` and you check the remaining elements count to break out of the loop? Couldn't you achieve only a single variable assignment that way?

packages/iterator-utilities/src/lib/max.ts
```diff
@@ -0,0 +1,31 @@
+import type { IterableResolvable } from './from';
+import { map } from './map';
+import { toNumberOrThrow, type NumberResolvable } from './shared/_toNumberOrThrow';
+
+/**
+ * Consumes the iterable and returns the maximum element. If the iterable is empty, it returns `null`.
```
favna, line 6:
```suggestion
 * Consumes the iterable and returns the highest number element. If the iterable is empty, or contains only non-number values, it returns `null`.
```

packages/iterator-utilities/src/lib/min.ts
```diff
@@ -0,0 +1,31 @@
+import type { IterableResolvable } from './from';
+import { map } from './map';
+import { toNumberOrThrow, type NumberResolvable } from './shared/_toNumberOrThrow';
+
+/**
+ * Consumes the iterable and returns the minimum element. If the iterable is empty, it returns `null`.
```
favna, line 6:
```suggestion
 * Consumes the iterable and returns the lowest number element. If the iterable is empty, or contains only non-number values, it returns `null`.
```

packages/iterator-utilities/src/lib/slice.ts
```diff
@@ -0,0 +1,119 @@
+import { drop } from './drop';
+import { dropLast } from './dropLast';
+import { empty } from './empty';
+import type { IterableResolvable } from './from';
+import { toIntegerOrInfinityOrThrow } from './shared/_toIntegerOrInfinityOrThrow';
+import { take } from './take';
+import { takeLast } from './takeLast';
+
+/**
+ * Creates an iterable with the elements from the `start` index to the `end` index (exclusive).
```
favna, line 10: Specify that just like Array#slice this does not modify the original input

packages/iterator-utilities/src/lib/starMap.ts
favna, line 1: The documentation makes it very unclear to me how this differs from `map`. I see the code difference but it's still unclear to me. This difference should be specified in _both_ files.

I'm also wondering if `starMap` is the best name for this function but I can only judge that when I understand how this differs from `map`.

packages/iterator-utilities/src/lib/takeWhile.ts
```diff
@@ -0,0 +1 @@
+export { filter as takeWhile } from './filter';
```
favna, line 1: Can't you just export this frm `filter` similar to how you exported other aliases?

packages/iterator-utilities/src/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+	"extends": "../../../tsconfig.base.json",
+	"compilerOptions": {
+		"lib": ["dom", "esnext"],
```
favna, line 4: Does this really need DOM? If not drop the entire field because the base is already esnext.

packages/iterator-utilities/README.md
```diff
@@ -0,0 +1,765 @@
...
+```ts
+import { append } from '@sapphire/iterator-utilities';
+const { append } = require('@sapphire/iterator-utilities');
+```
+
+Or you can import the utilities directly:
+
+```ts
+import { append } from '@sapphire/iterator-utilities/append';
+const { append } = require('@sapphire/iterator-utilities/append');
```
favna, line 60: Does Require syntax even support that? Pretty sure it doesn't...?

favna (changes requested):
sorrynotsorry

packages/iterator-utilities/README.md
kyranet: Yes, it does, paths aren't an ESM-only feature after all

vladfrangu: Yes, it does. In my screenshot it _would've_ worked if i wasn't in highlight where its all esm

iTerm2 - 2024-04-28 at 15 28 53

packages/iterator-utilities/src/lib/last.ts
kyranet: Iterables are eager and do not have a length property.

packages/iterator-utilities/src/lib/slice.ts
kyranet: This was misleading, it should be `Produces` instead of `Creates` because it consumes the iterable in most cases (only exceptions being `slice(iterable)`, `slice(iterable, 0)`, `slice(iterable, 1, 0)` (start >= end) and `slice(iterable, -2, -3)` (start >= end)).

Array#slice doesn't modify the input, but this one *may*.

packages/iterator-utilities/src/lib/shared/_assertFunction.ts
```diff
@@ -0,0 +1,7 @@
+export function assertFunction<Fn extends (...args: any[]) => any>(value: Fn): Fn {
```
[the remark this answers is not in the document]
kyranet: No, that'd add too many unnecessary generic types, which makes the code noisier. `any`, unlike `unknown`, generally has less issues when used for generics. Nothing is typed with `any` from this, by the way.

packages/iterator-utilities/src/lib/starMap.ts
kyranet: Since you brought it to my attention, I tweaked the function a little more.

`map` and `starMap` are very similar, both `map` after all, however, `starMap` takes an iterable of iterables (which can be an array of tuples) and calls the function with each inner iterable's values as the function's parameters. `map` calls the function with the value and the index by comparison.

Using `map`, you can achieve the implementation of `starMap` doing this (although obviously it's less optimal than the implementation, since it uses 2 callbacks instead of one):
```typescript
map(iterable, (value) => cb(...toIterableIterator(value));
```

favna: Add some of that to the tsdoc.

```ts

* @remark 
* 
* While very similar to {@link map}, `starMap` takes an iterable of iterables (which can be an array of tuples) and calls the function with each inner iterable's values as the function's parameters. {@link map} calls the function with the value and the index by comparison.

File: packages/iterator-utilities/src/tsconfig.json [hunk 10]
```diff
@@ -1,7 +1,7 @@
 {
 	"extends": "../../../tsconfig.base.json",
 	"compilerOptions": {
-		"lib": ["dom", "esnext"],
+		"lib": ["esnext"],
```
favna, line 4: As I said, you can then drop the `lib` field entirely, `esnext` is the base set by @sapphire/ts-config

favna (changes requested):

favna (dismissed):

packages/iterator-utilities/src/lib/filter.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export function filter<const ElementType, const FilteredType extends ElementType>(
+	iterable: IterableResolvable<ElementType>,
+	callbackFn: (element: ElementType, index: number) => element is FilteredType
+): IterableIterator<FilteredType>;
+export function filter<const ElementType>(
+	iterable: IterableResolvable<ElementType>,
+	callbackFn: (element: ElementType, index: number) => boolean
+): IterableIterator<ElementType>;
+
```
vladfrangu, line 29: In the words of our dragonite, whats with the space

vladfrangu (changes requested):

vladfrangu (approved):

packages/iterator-utilities/src/lib/filter.ts
favna: dragonite fuck yo shit up

favna (approved):

PR #747, opened 2024-05-17
Add `capitalizeFirstLetter`, `hasAtLeastOneKeyInObject`, and `snakeToCamelCase`

favna:
image


packages/utilities/src/lib/hasAtLeastOneKeyInObject.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Checks whether any of the {@link keys} are in the {@link object}
+ * @param obj - The object to check
+ * @param keys The keys to find in the object
+ * @returns `true` if at least one of the {@link keys} is in the {@link object}, `false` otherwise.
+ *
+ * @example
+ * ```typescript
+ * const obj = { a: 1, b: 2, c: 3 };
+ * console.log(hasAtLeastOneKeyInObject(obj, 'a')); // true
+ * console.log(hasAtLeastOneKeyInObject(obj, 'd')); // false
```
imranbarbhuiya, line 14:
```suggestion
 * console.log(hasAtLeastOneKeyInObject(obj, ['a'])); // true
 * console.log(hasAtLeastOneKeyInObject(obj, ['d'])); // false
```

packages/utilities/src/lib/hasAtLeastOneKeyInObject.ts
```diff
@@ -0,0 +1,22 @@
...
+ * ```
+ */
+export function hasAtLeastOneKeyInObject<T extends NonNullObject, K extends keyof T>(
+	obj: T,
+	keys: readonly K[]
```
imranbarbhuiya, line 19: since it's a  validator and also a type guard, so imo it shouldn't restrict u to pass `K[]` only

imranbarbhuiya (changes requested):

packages/utilities/src/lib/hasAtLeastOneKeyInObject.ts
```diff
@@ -0,0 +1,22 @@
+import { isNullish } from './isNullOrUndefined';
+import type { NonNullObject } from './types';
+
+/**
+ * Checks whether any of the {@link keys} are in the {@link object}
+ * @param obj - The object to check
```
vladfrangu, line 6: i doubt this link tag will work

packages/utilities/src/lib/hasAtLeastOneKeyInObject.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @example
+ * ```typescript
+ * const obj = { a: 1, b: 2, c: 3 };
+ * console.log(hasAtLeastOneKeyInObject(obj, ['a'])); // true
+ * console.log(hasAtLeastOneKeyInObject(obj, ['d'])); // false
+ * ```
+ */
+export function hasAtLeastOneKeyInObject<T extends NonNullObject, K extends keyof T>(
+	obj: T,
+	keys: readonly (string | number | symbol)[]
```
kyranet, line 19: You can use the alias from TypeScript:
```suggestion
	keys: readonly PropertyKey[]
```

packages/utilities/src/lib/snakeToCamelCase.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Transforms text from `snAkE_cASE` to `camelCase`.
+ * @param str - Text to transform
+ * @returns the input `st_R` as `stR`
+ *
+ * @example
+ * ```ts
+ * snakeToCamelCase('hello_world') // 'helloWorld'
+ * ```
+ */
+export function snakeToCamelCase(str: string) {
```
vladfrangu, line 11: Add in a kebab-case-to-camel-case too since this function seems to handle both

vladfrangu (changes requested):

packages/utilities/src/lib/snakeToCamelCase.ts
```diff
@@ -0,0 +1,13 @@
...
+	return str.toLowerCase().replace(/([-_][a-z])/g, (group: string) => group.toUpperCase().replace('-', '').replace('_', ''));
```
kyranet, line 12: You can do this:
```suggestion
	return str.toLowerCase().replace(/([-_][a-z])/g, (group: string) => group.slice(1).toUpperCase());
```

kyranet (changes requested):

packages/utilities/src/lib/snakeToCamelCase.ts
favna: Interesting. I didn't know that.

packages/utilities/src/lib/hasAtLeastOneKeyInObject.ts
favna: Mostly, just had to fix the `object` -> `obj`

kyranet (approved):

packages/utilities/src/lib/hasAtLeastOneKeyInObject.ts
```diff
@@ -0,0 +1,22 @@
+import { isNullish } from './isNullOrUndefined';
+import type { NonNullObject } from './types';
+
+/**
+ * Checks whether any of the {@link keys} are in the {@link obj}
+ * @param obj - The object to check
+ * @param keys The keys to find in the object
+ * @returns `true` if at least one of the {@link keys} is in the {@link object}, `false` otherwise.
```
vladfrangu, line 8: Ahem, object vs obj

vladfrangu (changes requested):

packages/utilities/src/lib/hasAtLeastOneKeyInObject.ts
```diff
@@ -0,0 +1,22 @@
...
+ * ```typescript
+ * const obj = { a: 1, b: 2, c: 3 };
+ * console.log(hasAtLeastOneKeyInObject(obj, ['a'])); // true
+ * console.log(hasAtLeastOneKeyInObject(obj, ['d'])); // false
+ * ```
+ */
+export function hasAtLeastOneKeyInObject<T extends NonNullObject, K extends keyof T>(
+	obj: T,
+	keys: readonly PropertyKey[]
+): obj is T & { [key in K]: NonNullable<T[K]> } {
```
imranbarbhuiya, line 20:
```suggestion
export function hasAtLeastOneKeyInObject<T extends NonNullObject, K extends PropertyKey>(
	obj: T,
	keys: readonly K[]
): obj is T & { [key in K]: NonNullable<T[K]> } {
```

now the typeguard will work properly, using `K extends PropertyKey & keyof T` allows any key and also gives auto complete for known keys but it breaks the typeguard, maybe u can experiment with it to make autocomplete and typeguard both work while also allowing any keys

imranbarbhuiya (changes requested):

kyranet (approved):

vladfrangu (approved):

PR #750, opened 2024-05-21
ci(scripts): add gen-index

### LINKED ISSUE #749: [SFW-82] request: script to automatically generate index.ts re-exports for /utilities

### Feature description

It's easy to forget to re-export the utilities on index.ts level. To solve this it would be prudent to add a script and make the file completely auto-generated at pre-build time.

### Desired solution

- A script that automatically injects the code and runs automatically before the code gets build
- A step in CI that validates if the file is correct

### Alternatives considered

suffer.exe

### Additional context

_No response_

<sub>[SFW-82](https://linear.app/sapphiredev/issue/SFW-82/request-script-to-automatically-generate-indexts-re-exports-for)</sub>

Nytelife26:
Fixes #749.

### Current notes

Note that for the purposes of the following ruleset, types and interfaces are treated equally.
1. If the module begins with `_`, ignore it
  a) unless the module contains types, in which case those exclusively should be included
2. If the module exclusively contains types, export with `export type *`, otherwise use `export *`

#### Consolidated list of packages to skip
- eslint-config
- eslint-plugin-result
- event-iterator (note: this is because the package is just an index)
- prettier-config
- stopwatch (note: this is because the package is just an index)
- time-utilities (note: this is because the package is a re-export of different packages)
- ts-config
- discord.js-utilities (note: this is because the package re-exports `@sapphire/discord-utilities`, which is temporary and would be fixed as described below)

#### Performance

Performance metrics on an i7-12700H under WSL for all applicable packages are:
image
Internal profiling reveals that from script loading to generation takes ~600-700ms. A further ~300ms per run is added by yarn, tsx and node bootstrapping. The difference can be seen in the following screenshot, where the script handles all packages in one run.
image
This serialised approach actually turns out to be faster than running the script individually via turbo (without cache) in parallel mode, demonstrating that startup is a huge part of the overhead. If there's an easy way to run the script on all changed packages at once, maybe implementing similar parallelisation, that would drastically reduce the impact on build times. This could potentially be achieved via lint-staged, or similar mechanisms.

### Future improvements

Ideally, it would be nice to have it leave external re-exports intact. This would require reading and processing the previous `index.ts`, which would slow things down, but it's all we need to safely include things like `time-utilities` and `discord.js-utilities`. As a workaround in the meantime, people may consider adding these re-exports to other files which are then exported by the index.

### Current disparities

- tries to export `AsyncQueueEntry` in `AsyncQueue`


favna:
> use export { ..., type ... } for both types and interfaces

This is indeed the way to go. We use types when they are direct aliases (which you can see for throttle) and interfaces when they represent objects (as is the case with isThenable)

To show that in code, the near equivalent of an interface for the type alias used in Throttle would be
```ts
export interface ThrottleFn<T extends (...args: any[]) => any> extends T {
	/**
	 * Resets the last time the throttled function was invoked.
	 */
	flush(): void;
}
```
However, this is invalid code, throwing the error
```
Interface 'ThrottleFn<T>' incorrectly extends interface 'T'.
  'ThrottleFn<T>' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint '(...args: any[]) => any'.ts(2430)
```

On to another topic, the PR #754 should alleviate some of the issues this PR is facing by making re-exports on utilities easier.

Looking at the code after #754, the only places that do not exclusively use `export * from 'path'` in `index.ts` now are:
- `eslint-config`: This package should not be parsed by this script to begin with
- `eslint-plugin-result`: Same as above
- `event-iterator`: Same as above because it only has 1 index.ts file
- `iterator-utilities`: This has a single `export type` from a `_` prefixed file. This was addressed in your comment.
- `phisherman`: This has a single `export type *` because the file only has types. This could become `export *` but if possible to detect if the file only has types/interfaces then `export type *` would be preferred.
- `prettier-config`: Same as `eslint-config`
- `stopwatch`: Same as `event-iterator`, the script should skip this package
- `time-utilities`: Mentioned above, it's a re-export of different packages. The script can skip this package.
- `ts-config`: Same as `eslint-config` and `prettier-config`

With "skip this package" what I actually mean is that nothing is modified in their respective `package.json` files to call the new script after building, by doing this and setting up a turbo pipeline on root it will automatically be "skipped". No code needs to be added to the script to account for these exclusions.

.github/workflows/continuous-integration.yml
```diff
@@ -118,3 +118,32 @@ jobs:
...
+      - name: Use Node.js v20
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: yarn
+          registry-url: https://registry.yarnpkg.org/
+      - name: Install Dependencies
+        run: yarn --immutable
+      # It would be nice to cache turbo's output, but it passes for all subsequent runs since it doesn't commit to fs
+      - name: Generate index files
```
favna, line 137:
```suggestion
      - name: Generate index files
```

I don't really like these kind of code comments. It's not useful to add.

favna (changes requested):
so.. what about the todos in the script?

scripts/gen-index.cts
```diff
@@ -0,0 +1,148 @@
...
+	 * Rules:
+	 * 1. If the module begins with `_`, ignore it
+	 *    a) unless the module contains types, in which case those exclusively should be included
+	 * 2. If the module exclusively contains types, use `export type *`, otherwise use `export *`
+	 **/
+	public toExportDeclaration(packageDir: string): ts.ExportDeclaration | undefined {
+		const { useModule, useNormal, useTypes } = this.exportInclusions;
+		if (!useModule) return;
+
+		// TODO: make this more efficient
```
favna, line 61: Resolve in this PR, or delete the TODO. Your call. Adding a random TODO is poor management of technical debt. Taking up tech. debt is something we only do if there is a good reason to do so.

scripts/gen-index.cts
```diff
@@ -0,0 +1,148 @@
...
+			useNormal: !this.isPrivate && this.exports.normal.length > 0,
+			useTypes: typesAvailable
+		};
+	}
+
+	private parseExports(): ModuleExportNodes {
+		const normal: ts.Declaration[] = [];
+		const types: ts.Declaration[] = [];
+
+		// TODO: skip normals where this.isPrivate?
```
favna, line 90: Resolve in this PR, or delete the TODO. Your call. Adding a random TODO is poor management of technical debt. Taking up tech. debt is something we only do if there is a good reason to do so.

scripts/gen-index.cts
```diff
@@ -0,0 +1,148 @@
...
+async function findIndexOrModules(dir: string, depth: number = 0): Promise<string[]> {
+	const contents = await readdir(dir, { withFileTypes: true });
+	let results: string[] = [];
+	for (const item of contents) {
+		if (item.name === 'common') continue;
+		const itemPath = join(dir, item.name);
+		if (item.isFile()) {
+			if (!item.name.endsWith('.ts')) continue;
+			if (item.name === 'index.ts') {
+				// TODO: process index to support re-exports (see #750)
```
favna, line 112: Resolve in this PR, or delete the TODO. Your call. Adding a random TODO is poor management of technical debt. Taking up tech. debt is something we only do if there is a good reason to do so.

scripts/gen-index.cts
```diff
@@ -0,0 +1,148 @@
...
+				if (depth === 0) continue;
+				results = [itemPath];
+				break;
+			}
+			results.push(itemPath);
+			// TODO: there are likely more efficient ways to determine if a directory has an auxiliary index
```
favna, line 118: Resolve in this PR, or delete the TODO. Your call. Adding a random TODO is poor management of technical debt. Taking up tech. debt is something we only do if there is a good reason to do so.

favna (changes requested):

.github/workflows/continuous-integration.yml
```diff
@@ -118,3 +118,31 @@ jobs:
...
+        run: yarn gen-index --parallel --concurrency $(ls packages | wc -l) --output-logs new-only
+      - name: Test for differences
+        run: |
+          git add .;
+          if git diff-index --quiet HEAD --; then
+            echo "No index discrepancies found, exiting with code 0";
+            exit 0;
+          else
+            echo "Index discrepancies found, displaying and exiting with code 1";
+            git diff;
```
Nytelife26, line 146: gotta fix this one since it doesn't display

scripts/gen-index.cts
Nytelife26: upon further analysis doing that would require calling `isType` anyway so there would be little benefit, it would merely avoid unnecessary allocations, which doesn't really matter in a small script. removing shortly

scripts/gen-index.cts
Nytelife26: that's a pretty hefty one, probably best handled in a separate pr, which i can do. if it's not a feature you're interested in, i could also remove it

favna: Just remove it tbh. I don't see this being needed any time soon.

.github/workflows/continuous-integration.yml
favna: Try writing it to a file then setting error similar as its done here: https://github.com/skyra-project/discord-components/blob/70263974e2950a9c0c67ca803546bd2fc880bf88/.github/workflows/generated-file-validations.yml#L31

Failed run example: https://github.com/skyra-project/discord-components/actions/runs/7861891265

scripts/gen-index.cts
Nytelife26: added it anyway, that'll make `discord.js-utilities` easier to maintain, and that also means if other packages ever need to re-export something they can leverage our index generation tool

PR #754, opened 2024-05-28
refactor(utilities): rename `baseVariants` to `toTitleCaseDiscordJsVariants`

favna:
This constant is exported after #750 but the name was kind of vague


Nytelife26:
all of these except the name change of `baseVariants` will be resolved by #750 anyway, so i would recommend rescoping this to just the name change. other than that, world's easiest lgtm

PR #758, opened 2024-06-03
ci: setup cliff-jumper v4 and publishing workflow

favna:


kyranet (dismissed):

.github/workflows/publish.yml
```diff
@@ -0,0 +1,51 @@
...
+    name: Publish @sapphire/${{ github.event.inputs.package}}
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'sapphiredev'
+    steps:
+      - name: Checkout Project
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: main
+          token: ${{ secrets.SKYRA_TOKEN }}
```
vladfrangu, line 21: Why don't you clone with the github_token, if you use it for literally everything else down below?

packages/async-queue/cliff.toml
```diff
@@ -24,7 +27,10 @@ body = """
 		- {% if commit.scope %}\
 			**{{commit.scope}}:** \
 		  {% endif %}\
-            {{ commit.message | upper_first }} ([{{ commit.id | truncate(length=7, end="") }}](https://github.com/sapphiredev/utilities/commit/{{ commit.id }}))\
+            {{ commit.message | upper_first }} ([{{ commit.id | truncate(length=7, end="") }}]({{ self::remote_url() }}/commit/{{ commit.id }}))\
+            {% if commit.github.pr_number %} (\
+                [#{{ commit.github.pr_number }}]({{ self::remote_url() }}/pull/{{ commit.github.pr_number }})) \
+            {%- endif %}\
```
vladfrangu, line 33: Can we get the PR authors in this now?

.github/workflows/publish.yml
favna: Because then you run into not being allowed to push to main: https://github.com/favware/cliff-jumper/actions/runs/9353623553/workflow

vladfrangu: But you literally change the origin URL to the other token?

packages/async-queue/cliff.toml
favna: yes, pushed in [f19ff26](https://github.com/sapphiredev/utilities/pull/758/commits/f19ff264c77dc9bd0fd5e002b61213c4292542d2)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Use Node.js v20
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: yarn
+          registry-url: https://registry.yarnpkg.com/
+      - name: Install Dependencies
+        run: yarn --immutable
```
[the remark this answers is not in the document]
favna: It never was, not for v3 and not even for v1 (as --frozen-lockfile). Yarn has always checked for `process.env.CI`. Doesn't take away that it's a good extra security guard to add.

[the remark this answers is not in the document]
vladfrangu: ...Yes, it does, and I have CI proof from work to show it. This is redundant

.github/workflows/publish.yml
favna: Doesn't take away that without the CI run fails as shown above. How the origin URL plays into that I don't know. I also tried removing setting the URL but that too failed. This is the way I've been running automatic updates for graphql-pokemon for a very long time now.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: yarn bump
+        working-directory: packages/${{ github.event.inputs.package }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Publish to NPM
+        working-directory: packages/${{ github.event.inputs.package }}
+        run: |
+          yarn config set npmAuthToken ${NODE_AUTH_TOKEN}
+          yarn config set npmPublishRegistry "https://registry.yarnpkg.com"
+          yarn npm publish
```
vladfrangu, line 49: Lets get provenances on too

vladfrangu (changes requested):

.github/workflows/publish.yml
favna: No can do, requires https://github.com/yarnpkg/berry/issues/5430

favna: And before you say "then use npm publish instead", no can do either because it would break dependencies such as `"@sapphire/utilities": "workspace:^"`

.github/workflows/publish.yml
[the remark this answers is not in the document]
favna: See https://v3.yarnpkg.com/cli/install#:~:text=or%20switching%20branches.-,If%20the,option%20is%20set%20(defaults%20to%20true%20on%20CI)%2C,-Yarn%20will%20abort:
image

And okay it was required for Yarn v1, I stand corrected. I haven't used Yarn v1 in forever though. 

image


Either way, we still provide the flag in every single sapphire repo in every single install command so that's cleanup for later

vladfrangu (approved):

PR #775, opened 2024-07-21
feat: add `DeepReadonly` type

kyranet:


vladfrangu (dismissed):

packages/iterator-utilities/package.json
```diff
@@ -687,7 +687,7 @@
 		}
 	},
 	"sideEffects": [
-		"./dist/esm/chunk-G5GHKT7C.mjs",
```
favna, line 690: Undo this change so it doesnt show up in the conventional commit scan as a feat for iterator-utilities

favna (changes requested):

packages/iterator-utilities/package.json
kyranet: Done

favna (approved):

packages/utilities/README.md
```diff
@@ -589,6 +590,21 @@ A union of all builtin types.
...
+
+```ts
+type Foo = Set<{ bar?: ['foo', { hello: 'world' }] }>;
+
+// ReadonlySet<{
+//     readonly bar?: readonly ["foo", {
+//         readonly hello: "world";
+//     }] | undefined;
+// }>
+declare const foo: DeepRequired<Foo>;
```
HigherOrderLogic, line 605: This should be `DeepReadonly` right?

kyranet: ... yes, nice catch, thank you!

kyranet: Opened #776, thanks!

PR #777, opened 2024-07-23
docs: add usage example in README

kyranet:


packages/async-queue/README.md
```diff
@@ -58,3 +95,7 @@ Thank you to all the people who already contributed to Sapphire!
 </a>
 
 [contributing]: https://github.com/sapphiredev/.github/blob/main/.github/CONTRIBUTING.md
+
+```
+
+```
```
favna, line 101: huh?

favna (changes requested):

packages/async-queue/README.md
kyranet: I'm just as confused as you are, idk where that comes from, I didn't write it

favna (approved):

PR #780, opened 2024-07-26
feat(utilities): expose `Arr` to the docs

favna:
Fixes this warning:

> [warning] `Arr`, defined in
> `./projects/utilities/packages/utilities/src/lib/types.ts`,
> is referenced by `AbstractCtor.A` but not included in the documentation

### DIFF
--- a/packages/utilities/src/lib/types.ts
+++ b/packages/utilities/src/lib/types.ts
@@ -60,7 +60,6 @@ export type ArgumentTypes<F extends (...args: any[]) => unknown> = F extends (..
 
 /**
  * A readonly array of any values.
- * @private
  */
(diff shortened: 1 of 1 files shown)


kyranet (commented):
I believe we should also rename the type to something more descriptive than just `Arr` before making it public

favna:
Considering the type was already exported on a TS level I think that would be a breaking change. Thoughts @vladfrangu ?

vladfrangu:
> Considering the type was already exported on a TS level I think that would be a breaking change. Thoughts @vladfrangu ?

if its a private type, i'd be willing to accept it as sem minor

kyranet (approved):
I'm pleased now

PR #792, opened 2024-08-25
perf(snowflake): optimize Snowflake.timestampFrom

advaith1:
updates `timestampFrom` to operate on numbers instead of BigInts. This is faster whether the provided id is a string or bigint, even though sapphire stores the epoch as a bigint, and it is always accurate.

image
image

### DIFF
--- a/packages/snowflake/src/lib/Snowflake.ts
+++ b/packages/snowflake/src/lib/Snowflake.ts
@@ -173,7 +173,7 @@ export class Snowflake {
 	 * @returns The UNIX timestamp that is stored in `id`.
 	 */
 	public timestampFrom(id: string | bigint): number {
-		return Number((BigInt(id) >> 22n) + this[EpochSymbol]);
+		return Math.floor(Number(id) / 2 ** 22) + Number(this[EpochSymbol]);
 	}
 
 	/**
--- a/packages/snowflake/src/lib/Snowflake.ts
+++ b/packages/snowflake/src/lib/Snowflake.ts
@@ -1,5 +1,6 @@
 const IncrementSymbol = Symbol('@sapphire/snowflake.increment');
 const EpochSymbol = Symbol('@sapphire/snowflake.epoch');
+const EpochNumberSymbol = Symbol('@sapphire/snowflake.epoch.number');
 const ProcessIdSymbol = Symbol('@sapphire/snowflake.processId');
 const WorkerIdSymbol = Symbol('@sapphire/snowflake.workerId');
 
@@ -17,6 +18,8 @@ export const MaximumProcessId = 0b11111n;
  * The maximum value the `increment` field accepts in snowflakes.
  */
 export const MaximumIncrement = 0b111111111111n;
+
+const TimestampFieldDivisor = 2 ** 22;
 
 /**
  * A class for generating and deconstructing Twitter snowflakes.
@@ -45,6 +48,12 @@ export class Snowflake {
 	private readonly [EpochSymbol]: bigint;
 
 	/**
+	 * Internal reference of the epoch passed in the constructor as a number
+	 * @internal
+	 */
+	private readonly [EpochNumberSymbol]: number;
+
+	/**
 	 * Internal incrementor for generating snowflakes
 	 * @internal
 	 */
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

packages/snowflake/src/lib/Snowflake.ts
```diff
@@ -173,7 +173,7 @@ export class Snowflake {
 	 * @returns The UNIX timestamp that is stored in `id`.
 	 */
 	public timestampFrom(id: string | bigint): number {
-		return Number((BigInt(id) >> 22n) + this[EpochSymbol]);
+		return Math.floor(Number(id) / 2 ** 22) + Number(this[EpochSymbol]);
```
kyranet, line 176: Two suggestions:

- Regarding `2 ** 22`:
  - What’s the performance if we replace it with `1 << 22`?
  - What if we made it a constant?
- Regarding `Number(this[EpochSymbol])`:
  - What if we make `EpochNumberSymbol` which is the result as a number?

kyranet (changes requested):
This is clever, whilst JS numbers only have 53 bits of precision, the timestamp part is in the top 42 bits, so this will always work. I knew this could be optimised but I couldn’t think of a solution, this is it, thank you very much!

Just a couple things, although it’ll be better to benchmark this with vitest rather than the browser, that’ll probably give more accurate results.

packages/snowflake/src/lib/Snowflake.ts
vladfrangu: Fastest seems to be caching the power of 2 multiplication, and having numeric epochs

```
┌─────────┬────────────────────────────────────────────────────┬──────────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name                                          │ ops/sec      │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼────────────────────────────────────────────────────┼──────────────┼────────────────────┼──────────┼─────────┤
│ 0       │ 'bigint shift + numeric conversion'                │ '9,620,374'  │ 103.94605202702998 │ '±1.80%' │ 962038  │
│ 1       │ 'division with cached power of 2 and epoch'        │ '26,279,831' │ 38.0519946088112   │ '±0.58%' │ 2627984 │
│ 2       │ 'division with cached left shift and epoch'        │ '25,828,721' │ 38.716589240114665 │ '±0.59%' │ 2582873 │
│ 3       │ 'division with power of 2 and epoch'               │ '25,573,364' │ 39.1031850708978   │ '±0.67%' │ 2557337 │
│ 4       │ 'division with left shift and epoch'               │ '25,703,634' │ 38.90500372692088  │ '±0.87%' │ 2570364 │
│ 5       │ 'division with cached power of 2 and bigint epoch' │ '16,038,767' │ 62.34892887667072  │ '±6.64%' │ 1603877 │
│ 6       │ 'numeric left shift and epoch'                     │ '25,424,629' │ 39.33193980271842  │ '±1.07%' │ 2542464 │
│ 7       │ 'numeric left shift and bigint epoch'              │ '21,102,698' │ 47.3873035200918   │ '±0.58%' │ 2110270 │
└─────────┴────────────────────────────────────────────────────┴──────────────┴────────────────────┴──────────┴─────────┘
```

<details>
<summary>Test Code</summary>

```ts
import { Bench } from 'tinybench';

const bench = new Bench({ time: 100 });

const id = '139836912335716352';
const discordEpoch = 1_420_070_400_000n;
const numericDiscordEpoch = 1_420_070_400_000;
const powerOfTwoCached = 2 ** 22;
const leftShiftCached = 1 << 22;

bench.add('bigint shift + numeric conversion', () => {
	Number((BigInt(id) >> 22n) + discordEpoch);
});

bench.add('division with cached power of 2 and epoch', () => {
	Math.floor(Number(Number(id) / powerOfTwoCached + numericDiscordEpoch));
});

bench.add('division with cached left shift and epoch', () => {
	Math.floor(Number(Number(id) / leftShiftCached + numericDiscordEpoch));
});

bench.add('division with power of 2 and epoch', () => {
	Math.floor(Number(Number(id) / 2 ** 22 + numericDiscordEpoch));
});

bench.add('division with left shift and epoch', () => {
	Math.floor(Number(Number(id) / (1 << 22) + numericDiscordEpoch));
});

bench.add('division with cached power of 2 and bigint epoch', () => {
	Math.floor(Number(Number(id) / powerOfTwoCached + Number(discordEpoch)));
});

bench.add('numeric left shift and epoch', () => {
	Math.floor(Number((Number(id) >> 22) + numericDiscordEpoch));
});

bench.add('numeric left shift and bigint epoch', () => {
	Math.floor(Number((Number(id) >> 22) + Number(discordEpoch)));
});

await bench.warmup();
await bench.run();

console.table(bench.table());
```

</details>

vladfrangu: One thing to keep in mind: numeric left shfits *lose* precision! They cannot be used, I just threw them in for completeness sake

vladfrangu (changes requested):
TL;DR:

- maybe cache the 2 ** 22 result in a constant
- create a `this[EpochNumericSymbol]` that is set on creation to the numeric value of the epoch and use it

favna (approved):

kyranet (approved):

vladfrangu (approved):

PR #794, opened 2024-08-27
feat: clonable TwemojiRegex

vladfrangu:


favna (changes requested):
You also need to update https://github.com/sapphiredev/utilities/blob/main/scripts/twemoji-regex-updater.cts otherwise https://github.com/sapphiredev/utilities/blob/main/.github/workflows/auto-updater.yml will nuke this change on next run

favna (approved):

PR #797, opened 2024-09-04
refactor(discord-utilities): switch to jdecked twemoji fork

SerenModz21:
Discord has switched to using Justine De Caires' (jdecked) Twemoji fork. Justine is an ex-Twitter employee who off the top of my head, used to work on Twemoji but got laid off during the layoffs after Elon bought Twitter/X. Therefore, switching Sapphire's regex to using said fork would be better suited.

Benefits of this change:
- Adds support for Twemoji v15
- Uses the same version as Discord
- More updated than the original Twitter repo

Sources:
- https://x.com/DiscordPreviews/status/1640809457423732763
  Discord Previews mentioned the switch back on the 28th of March 2023.
- https://github.com/discord/twemoji
  Discord's Twemoji repo is a fork of Justine's fork.


vladfrangu (changes requested):
Please also update https://github.com/sapphiredev/utilities/blob/main/scripts/twemoji-regex-updater.mts so it fetches from the new source!

SerenModz21:
Thank you for your review Vlad! The script was the first thing that I changed and ran it afterwards to update the actual code, since it gets overridden anyway.

https://github.com/sapphiredev/utilities/pull/797/files/8f4b64fe0bbd4194159acc88f0889f6141ce32e2#diff-ea04fd7ec7f44a64238cca194037a694d90415d43bb2f5a871379f49544fd852

vladfrangu (approved):
Not me being blind

packages/iterator-utilities/package.json
```diff
@@ -687,7 +687,7 @@
 		}
 	},
 	"sideEffects": [
-		"./dist/esm/chunk-G5GHKT7C.mjs",
+		"./dist/esm/chunk-PAWJFY3S.mjs",
```
favna, line 690:
```suggestion
		"./dist/esm/chunk-G5GHKT7C.mjs",
```

This needs to be reverted otherwise changelog generation will put the commit to discord-utilities on iterator-utilities as well

favna (changes requested):

vladfrangu (approved):

favna (approved):

favna (approved):

PR #799, opened 2024-09-08
fix: update for latest discord.js

favna:


package.json
```diff
@@ -35,8 +35,8 @@
 		"@vitest/coverage-v8": "^2.0.5",
 		"concurrently": "^8.2.2",
 		"cz-conventional-changelog": "^3.3.0",
-		"discord-api-types": "^0.37.98",
-		"discord.js": "^14.15.3",
+		"discord-api-types": "^0.37.100",
```
vladfrangu, line 38: We really ought to keep this tied to the djs version

vladfrangu (dismissed):

package.json
vladfrangu: That sounds like an oversight on our part, not intentional

favna: Pinned to .97 and mentioned in djs release talk chat

PR #806, opened 2024-09-29
fix: resolve build errors

kyranet:


packages/result/src/lib/Option/None.ts
```diff
@@ -1,17 +1,17 @@
...
+import type { Option, Some } from '../Option';
 import { createErr, type ResultErr } from '../Result/Err';
 import { createOk, type ResultOk } from '../Result/Ok';
 import type { IOption } from './IOption';
 import { OptionError } from './OptionError';
 import type { OptionSome } from './Some';
 
 export class OptionNone implements IOption<any> {
-	public isSome(): false {
+	public isSome(): this is Some<any> & false {
```
vladfrangu, line 10: Remove the `& false`, its useless

packages/result/src/lib/Option/None.ts
```diff
@@ -108,7 +108,7 @@ export class OptionNone implements IOption<any> {
 	}
 
 	public filter(predicate: (value: never) => boolean): OptionNone;
```
vladfrangu, line 110: As well as all other places

```suggestion
	public filter(predicate: (value: never) => boolean): this;
```

vladfrangu (changes requested):

packages/result/src/lib/Option/None.ts
```diff
@@ -7,11 +7,11 @@ import { OptionError } from './OptionError';
...
 
 export class OptionNone implements IOption<any> {
-	public isSome(): false {
+	public isSome(): this is OptionSome<any> {
 		return false;
 	}
 
-	public isSomeAnd(cb?: (value: never) => boolean): false;
+	public isSomeAnd(cb?: (value: never) => boolean): this is OptionSome<any>;
 	public isSomeAnd(): false {
```
vladfrangu, line 15: This should also be a typeguard, no?

packages/iterator-utilities/package.json
```diff
@@ -687,7 +687,7 @@
 		}
 	},
 	"sideEffects": [
-		"./dist/esm/chunk-G5GHKT7C.mjs",
+		"./dist/esm/chunk-PAWJFY3S.mjs",
```
favna, line 690: revert

favna (changes requested):

packages/result/src/lib/Option/None.ts
kyranet: Unneeded, this method isn't overloaded.

vladfrangu (approved):

favna (approved):

PR #808, opened 2024-09-29
feat: rewrite `/result`

kyranet:
This update brings a simplification to the `result` package, removing the duplication of the `Option` and `Result` types and their respective implementations. With this new approach, the `Option` and `Result` types now have a single implementation, which is more in line with the Rust language.

No breaking changes were introduced, and the tests were updated to reflect the new implementation, however, the types have changed drastically and not longer have the same structure as before. Among the changes, a lot of overloads were removed in favour of a single implementation that uses generics to handle the different cases.


packages/result/src/lib/Option.ts
```diff
@@ -1,50 +1,935 @@
...
+	 * ```typescript
+	 * await none.inspectAsync(console.log);
+	 * // Doesn't log
+	 * ```
+	 *
+	 * @note This is an extension not supported in Rust
+	 */
+	public async inspectAsync(cb: (value: T) => Awaitable<unknown>): Promise<this> {
+		if (this.isSome()) await cb(this[ValueProperty]);
+		return Promise.resolve(this);
```
vladfrangu, line 343: direct return please, no need to re-create a promise

packages/result/src/lib/Option.ts
```diff
@@ -1,50 +1,935 @@
...
+	 * ```
+	 *
+	 * @see {@link https://doc.rust-lang.org/std/option/enum.Option.html#method.unzip}
+	 */
+	public unzip<Value0, Value1, Exists extends boolean>(
+		this: Option<readonly [Value0, Value1], Exists>
+	): [Option<Value0, Exists>, Option<Value1, Exists>] {
+		// @ts-expect-error Complex types
+		return this.match({
+			some: ([value0, value1]) => [some(value0), some(value1)],
```
vladfrangu, line 711: you don't check anywhere that `this` is a tuple/zipped up option

vladfrangu (dismissed):

packages/result/src/lib/Option.ts
kyranet: Nor did the original code:
https://github.com/sapphiredev/utilities/blob/809c41cef869bb9095cd630a416981be6870e83a/packages/result/src/lib/Option/Some.ts#L142-L145

It's "checked" with the `this` guard, which hides this method altogether if you try to use it on an incompatible `Option`.

favna (changes requested):
CI is failing

favna (approved):

favna (approved):

PR #809, opened 2024-10-01
feat(iterator-utilities): add more methods

kyranet:
Added the following functions:

- `all` (alias for `every`)
- `any` (alias for `some`)
- `collect` (alias for `toArray`)
- `collectInto`
- `compare`
- `compareBy`
- `equal`
- `equalBy`
- `fuse`
- `greaterOrEqualThan` (with alias for `ge`)
- `greaterThan` (with alias for `gt`)
- `inspect`
- `intersperse`
- `isSorted`
- `isSortedBy`
- `isSortedByKey`
- `lessOrEqualThan` (with alias for `le`)
- `lessThan` (with alias for `lt`)
- `maxBy`
- `maxByKey`
- `minBy`
- `minByKey`
- `notEqual`
- `nth` (alias for `at`)
- `stepBy`
- 🔥 **BREAKING CHANGE**: `min` now uses a lexicographical comparison by default. Use `minBy` for custom comparisons.
- 🔥 **BREAKING CHANGE**: `max` now uses a lexicographical comparison by default. Use `maxBy` for custom comparisons.


packages/iterator-utilities/src/lib/equalBy.ts
```diff
@@ -0,0 +1,44 @@
...
+	other: IterableResolvable<ElementType>,
+	callbackFn: (x: ElementType, y: ElementType) => boolean
+): boolean {
+	callbackFn = assertFunction(callbackFn);
+
+	const iterator1 = from(other);
+
+	for (const value0 of toIterableIterator(iterable)) {
+		const result1 = iterator1.next();
+		if (result1.done || !callbackFn(value0, result1.value)) return false;
```
vladfrangu, line 38: Can't we get a fastpath for arrays?

kyranet: Quite a lot of things could get a fastpath for arrays, but that'll require adding more code to handle those cases.

That being said, I considered the possibility of having a `sizeHint` in our iterator.

favna (approved):

PR #814, opened 2024-10-19
fix(fetch): support stringifying arrays

favna:


packages/fetch/src/lib/fetch.ts
```diff
@@ -202,6 +202,8 @@ function shouldJsonStringify(value: unknown) {
 	if (value.constructor === Object) return true;
 	// Has toJSON method
 	if ('toJSON' in value && typeof value.toJSON === 'function') return true;
+	// Array of items, check if every item in the array is serializable
+	if (Array.isArray(value)) return value.every(shouldJsonStringify);
```
vladfrangu, line 206: .map not .every

vladfrangu (dismissed):

kyranet (approved):

PR #820, opened 2024-10-27
refactor(cutText): improve algorithm

kyranet:


packages/utilities/src/lib/cutText.ts
```diff
@@ -1,4 +1,4 @@
-import { splitText } from './splitText';
+const WordSeparatorCharacter = /[\p{Separator}\p{Punctuation}\p{Control}]/u;
```
favna, line 1: What is the minimum Node version for `\p` and `u` flag in Regex? sapphire utilities targets Node >= 14.0.0 through the engines field.

favna: `u` is okay: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode

kyranet: Widely available, I checked on https://caniuse.com/mdn-javascript_builtins_regexp_unicode

PR #821, opened 2024-10-27
refactor: deprecate `NonNullObject` and `AnyObject`

kyranet:


favna (dismissed):
I'm pretty sure that I have had need of these at work where simply `object` didn't suffice or wasn't strictly typed enough but it's whatever I guess, I can always copy them over if needed.

Also linting fails

favna (approved):

PR #836, opened 2024-11-22
feat: add `windows`

kyranet:


packages/iterator-utilities/src/lib/windows.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @example
+ * ```typescript
+ * import { windows } from '@sapphire/iterator-utilities';
+ *
+ * const iterable = [1, 2, 3, 4, 5];
+ * console.log([...windows(iterable, 2)]);
+ * // Output: [[1, 2], [2, 3], [3, 4], [4, 5]]
+ * ```
+ */
+export function windows<const ElementType>(iterable: IterableResolvable<ElementType>, count: number): IterableIterator<ElementType[]> {
```
vladfrangu, line 22: Funny pun but shouldn't this be called `window` not `windows`?

kyranet: No, it's a method that makes windows, not a single window.

Also the name is based on the proposal: https://github.com/tc39/proposal-iterator-chunking

packages/iterator-utilities/tests/windows.test.ts
```diff
@@ -0,0 +1,46 @@
...
+	});
+
+	test('GIVEN empty iterable THEN returns empty iterable', () => {
+		const iterable: number[] = [];
+		const limit = 5;
+		const result = [...windows(iterable, limit)];
+		expect(result).toEqual([]);
+	});
+
+	test('GIVEN iterable and limit equal to 0 THEN returns empty iterable', () => {
```
vladfrangu, line 35: wrong test msg

vladfrangu (dismissed):

favna (approved):
Where are Linux and MacOS?

PR #472, opened 2022-09-30
feat(utilities): add `objectKeys` function

RealShadowNova:
Resolves #457

PR #471, opened 2022-09-30
feat(utilities): add `objectEntries` function

RealShadowNova:
Resolves #455

PR #857, opened 2025-01-18
fix(utilities): ensure consistent typing for `objectEntries`, `objectKeys` and `objectValues`

GeniusTimo:
I noticed that these functions have return types that do not match the type assertions within their implementations. After reviewing the pull requests that introduced these functions, I believe the idea to use `ArrayLike<infer Values>` (see https://github.com/sapphiredev/utilities/pull/474#discussion_r985112679) may have unintentionally caused some confusion. The idea for the `objectValues` function was only implemented to the type assertion within the function and not yet for its return type (de87080f4c045919c1676f0fb4f76102a95a5ab3 ✔)

To improve consistency, the same idea was adapted for the `objectEntries` function (see https://github.com/sapphiredev/utilities/pull/471#discussion_r985113263). However, it was implemented as if `Object.entries` returned the values, not the entries, when dealing with array-like objects (9d6d3e80ae7f66da732e44f99dad140f68967a41 ✔)
Because the return type of the function was not updated to reflect this change, it had no real impact. To align with the intended consistency, I have updated the return type to match the corrected type assertion (34c3205fb6a6f1780a0c6271c3b40703d82be6c8 ✔)

The type of the values within an array-like object is not relevant to the `objectKeys` function. However, the idea was also applied to that type assertion (see https://github.com/sapphiredev/utilities/pull/472#discussion_r985113543) Because the return type was not adjusted accordingly, this change also had no actual effect. I have fixed this as well (b43fba072b9e1f64f7472578f45533a4c4144063 ✔)

This pull request is meant to reduce potential confusion for future contributors. The details regarding the previous implementation are provided for traceability only and are not intended as criticism of or offense to anyone involved back then.


kyranet (approved):

favna (changes requested):
Could you add unit tests for this using [vitest testing types](https://vitest.dev/guide/testing-types.html)? To ensure that the same mistake is not made in the future because these types are quite complex.

GeniusTimo:
I'm not familiar with vitest yet, but I've tried my best. Since we can't test the type assertion within the function (at least to my knowledge), I feel like checking the return type for arrays is all we can add?

